PR #3, opened 2014-10-14
Handle shortcode rendering

mattheu:
Should the scope of this plugin extend to handling add_shortcode and the callback?
- Big + is that we can use the same templates internally.
- Bad - Not such good compat with existing shortcodes.

Could make this optional - eg if render template is provided then can specify render callback? What if they add shortcode too? maybe modify `$shortcode_tags[$tag]` directly?

I would like to keep options to a minimum and would like to have a certain ammount of backwards compat if a user chooses to move away from this plugin.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -45,6 +45,8 @@ class Shortcode_UI {
 
 		$args['shortcode'] = $shortcode;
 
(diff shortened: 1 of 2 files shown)


mattheu:
Merging - lets cross this bridge when we come to it - or something like that

PR #21, opened 2014-10-24
Tweaks from Daniel's code review

### LINKED ISSUE #11: get_view() shouldn't echo
I think it's a bad pattern to have `get_*` methods echo. Let's have it always return, and be consistent with echo'ing its return value elsewhere.

### LINKED ISSUE #15: Shortcode preview needs a nonce
```
$.post( ajaxurl, data, function( response ) {
                        // Note - set even if empty to prevent this firing multiple times.
                        t.shortcodeHTML = response;
                        t.render( true );
                    });
```

danielbachhuber:
- Nonce / caps check for shortcode previews. Fixes #15
- Mark methods as public; rename some to be more explicit.
- Fix odd double nature of `get_view()`. Fixes #11
- Proper escaping of translation functions.
- Formatting

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -32,16 +32,16 @@ class Shortcode_UI {
 		} );
 
 		add_action( 'media_buttons',         array( $this, 'action_media_buttons' ) );
-		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@mattheu give this a look when you have a chance and let me know if you have any thoughts on it.

mattheu:
Thanks @danielbachhuber - all looks good

PR #23, opened 2014-10-26
add stripslashes to fix broken AJAX

mattheu:
review & merge

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -180,7 +180,7 @@ class Shortcode_UI {
 	 */
 	public function handle_ajax_do_shortcode( ) {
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
What's the test case?

mattheu:
Test case: ajax request for shortcode 
`[blockquote align="left" source="test source"]Test quote[/blockquote]`

Data available in ajax callback is:
 `[blockquote align=\"left\" source=\"test source\"]Test quote[/blockquote]`

When calling `sanitize_text_field` on this we get: 
`[blockquote align=\\"left\\" source=\\"test source\\"]Test quote[/blockquote]`

Which then breaks the attribute args passed to the shortcode render function.

PR #35, opened 2014-11-05
Add two actions for before / after shortcode render

danielbachhuber:
If a theme or plugin needs to include extra HTML, this is the way to do
it


chrisvanpatten:
This would be pretty awesome. Working on integrating JuxtaposeJS and it'd be nice to have a way to enqueue their dependencies so the content renders properly in the preview/editor.

Issue #28, opened 2014-10-27
Should we inject the shortcode preview within an Iframe?


danielbachhuber:
This is what core does to sandbox it. It would work for us too if we inject the editor stylesheet.

danielbachhuber:
Apparently TinyMCE views does this automatically when the content contains a script tag: https://core.trac.wordpress.org/changeset/29179

PR #36, opened 2014-11-05
Duplicate core's approach of wrapping script embeds in iframe

danielbachhuber:
See #28

### DIFF
--- a/css/shortcode-ui-editor-styles.css.map
+++ b/css/shortcode-ui-editor-styles.css.map
@@ -3,5 +3,5 @@
   "file": "",
   "sources": ["sass/shortcode-ui-editor-styles.scss"],
   "names": [],
-  "mappings": "AAAA;EACC,YAAY"
+  "mappings": "AAAA;EACC,YAAY;EAEb,YAAY,0BAA0B;IAGnC,SAAS"
 }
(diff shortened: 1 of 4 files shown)


danielbachhuber:
Boomski:

image

css/shortcode-ui-editor-styles.css.map
```diff
@@ -3,5 +3,5 @@
   "file": "",
   "sources": ["sass/shortcode-ui-editor-styles.scss"],
   "names": [],
-  "mappings": "AAAA;EACC,YAAY"
+  "mappings": "AAAA;EACC,YAAY;EAEb,YAAY,0BAA0B;IAGnC,SAAS"
```
davisshaver, line 6: What's this?

davisshaver:
@danielbachhuber, this looks good to me. Only concern I have is replicating core functionality. Could you explain why we're doing this?

danielbachhuber:
> Only concern I have is replicating core functionality. Could you explain why we're doing this?

To clarify, here's core's existing functionality: https://core.trac.wordpress.org/browser/tags/4.0/src/wp-includes/js/mce-view.js#L126

However, the `setIframes` function is more closely coupled to the rest of the model, including fetching the preview and rendering. Our model for fetching the preview and rendering is separate, which means we'd have to refactor some to make it work.

The API is marked as experimental though, so I don't want to create a dependency that's going to break. In short, let's internalize for now so we can better ensure stability. It can be refactored later if it makes it into core.

css/shortcode-ui-editor-styles.css.map
danielbachhuber, line 6: Junk stuff

PR #44, opened 2014-11-11
Fieldmanager

mattheu:
**NOTE:** This is still in progress.

Integration with fieldnamager to build out fields.

Here's where i'm at so far - and my thoughts on this.

Added fieldnamager firlds for textarea, text and media.

**The good:** 

We have a funcitoning fields - including image upload field.

I have made the core code for displaying fields more customisable - allowing developers to specify both a template for each field, and also a custom view object - which can be used to build more complex field types such as media uploader. This means the fieldmanager code can live as a compeltely separate 'plugin'.

The basic fieldmanager fields worked really well just by passing the underscore template placeholders as label, value and ID.

*\* The Bad**
More complex fields like the media uploader don't really work so well out of the box. Media for example requires a few workarounds to make th JS work correctly, and update the values.

Example 1: select field. We don't have access to options when building the template. So it requries custom JS to add these.

Example 2:  media field preview image HTML isn't output in the template because we have no value at this point. I have hacked around this by using ajax to get the image src and replicating the markup required - but its hacky.

My concerns here are we do not get enough benefit from fieldmanager. If custom code is required for  each of the more complex fields - is it that much easier than building our own? Basically fieldmanager is PHP and Shortcake is JS and its hacky to mesh the two. 

Also by going too far down this route are we making fieldmanager a dependency which would put people off using Shortcake?

Any thoughts?

### DIFF
--- a/dev.php
+++ b/dev.php
@@ -9,12 +9,19 @@ add_action( 'init', function() {
 	add_shortcode( 'pullquote', function( $attr, $content = '' ) {
 
 		$attr = wp_parse_args( $attr, array(
-			'source' => ''
+			'source' => '',
+			'image'  => '',
 		) );
 
 		?>
 
-		<section class="pullquote">
+		<section class="pullquote" style="overflow: hidden;">
+			<?php if ( $attr['image'] ) : ?>
+			<div style="float: left; margin-right: 20px;">
+				<?php echo wp_get_attachment_image( $attr['image'], array( 112, 112 ) ); ?>
+			</div>
+			<?php endif; ?>
+
 			<?php echo esc_html( $content ); ?><br/>
 			<cite><em><?php echo esc_html( $attr['source'] ); ?></em></cite>
 		</section>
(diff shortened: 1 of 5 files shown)


mattheu:
That said - all of the optimisations to the core code here is definitely worth keeping

danielbachhuber:
> Also by going too far down this route are we making fieldmanager a dependency which would put people off using Shortcake?

My thinking is that we could eventually support custom field builders. We're starting with Fieldmanager because this is what Fusion uses, and we should solve it as much it solves our immediate needs.

joshlevinson:
It'd be cool to see if @jtsternberg would go for integration with https://github.com/WebDevStudios/CMB2

jtsternberg:
I built a similarish library which integrates CMB2: https://github.com/jtsternberg/Shortcode_Button
I downloaded this (shortcake) and tried playing w/ it, but looks fairly limited, and the in-line preview didn't work. I'm not sure what I'm missing, but hesitant to contribute if there is not an obvious advantage over https://github.com/jtsternberg/Shortcode_Button
If there IS, I'd be happy to work together to combine the goodness of both (rather than more fracturing).

mattheu:
Interesting - hadn't seen this. I guess its somewhat similar.

PR #48, opened 2014-11-12
Placeholder attribute

dbarlett:
Adds optional placeholder attribute to text, textarea, url, email, number, and date inputs.

### DIFF
--- a/inc/templates/edit-form.tpl.php
+++ b/inc/templates/edit-form.tpl.php
@@ -16,21 +16,21 @@
 <script type="text/html" id="tmpl-shortcode-ui-field-text">
 	<p class="field-block">
 		<label for="{{ data.attr }}">{{ data.label }}</label>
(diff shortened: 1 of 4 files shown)


inc/templates/edit-form.tpl.php
```diff
@@ -16,21 +16,21 @@
 <script type="text/html" id="tmpl-shortcode-ui-field-text">
 	<p class="field-block">
 		<label for="{{ data.attr }}">{{ data.label }}</label>
-		<input type="text" name="{{ data.attr }}" id="{{ data.attr }}" value="{{ data.value }}"/>
+		<input type="text" name="{{ data.attr }}" id="{{ data.attr }}" value="{{ data.value }}"<# if ( data.placeholder ){ print( ' placeholder="' + data.placeholder + '"' ); } #>/>
```
danielbachhuber: For all of these, can we just have `data.placeholder` default to an empty value to avoid the conditional? I'm not sure the element is appropriately escaped in this pattern.

danielbachhuber:
@dbarlett great idea! one minor tweak to make.

dbarlett:
@danielbachhuber good point, will update PR.

danielbachhuber:
:+1:

Issue #50, opened 2014-11-13
It seems that we need to have all atts for re editing a shortcode


chdenat:
My UC : I have already in my post a shortcode , let's say [Myshortcode] ...[/Myshortcode]
I've declared 2  additionals parameters inside my shortcode_ui_register_for_shortcode declaration , le 'color', 'size'.

If the shortcode is [Myshortcode color='....' size='....'] ...[/Myshortcode] I can open the edit window
If the shortcode is [Myshortcode] ...[/Myshortcode] I cannot open the edit window

It seems that all attributes  are mandatory for shortcake ... that is not the case IRL  as some are defaulted by contruct...

danielbachhuber:
cc @mattheu

danielbachhuber:
Fixed by #50

PR #54, opened 2014-11-18
Existing shortcodes without attributes should work

roborourke:
In issue #50 pre-existing shortcodes weren't matching the regex, take the following:

```
[pullquote]4 score and 7...[/pullquote]
```

The shortcode tag was matched as `pullquote]4`.

I made the match for non white space and no closing square brackets. Also the test for `matches[2]` was always coming back true as you can't use the `!=` operator with undefined.


mattheu:
Thanks!

danielbachhuber:
@mattheu we should have some test coverage for the JavaScripts...

Issue #60, opened 2014-11-24
Shortcode Attributes Not Rendered in TinyMCE


zebulonj:
I'm seeing that the Shortcake view (preview) in Tiny MCE includes the shortcode "content," but is missing the shortcode attributes.

Using a clean development environment ([VIP Quickstart](https://github.com/Automattic/vip-quickstart)), I added the bleeding edge version of Shortcake, as well as the [demo plugin](https://gist.github.com/Mte90/cb8a0e37565d219062ec). I then added a "Pullquote" content item via the media interface. Returning to the post, the pullquote content showed in the embedded preview, but the source (citation) did not. When previewing the post outside the admin, the shortcode renders fine.

See attached screenshots.
![screen shot 2014-11-23 at 9 17 50 pm](https://cloud.githubusercontent.com/assets/1703671/5161176/5d24df44-7356-11e4-9069-848aadc41386.png)
![screen shot 2014-11-23 at 9 17 56 pm](https://cloud.githubusercontent.com/assets/1703671/5161177/5d26154e-7356-11e4-8deb-8e88b3899a6e.png)
![screen shot 2014-11-23 at 9 18 21 pm](https://cloud.githubusercontent.com/assets/1703671/5161178/5d2b16b6-7356-11e4-9d27-21d3ae4f7f56.png)

I did a bit of diagnostic work, and it appears that when fetched in the admin via the ajax method, the shortcode attributes are not being parsed correctly via `wp_parse_args(...)`. For example, the string 'source="The Dude"' is being returned as the value at index 0 in the associative array.

I'm happy to dig deeper, but was hoping someone else could verify this behavior before I do... in case it's my setup.

Thanks!

roborourke:
I'm seeing this too with the wordpress-develop repo.

danielbachhuber:
Caused by 3ad09a2459f1e81f6f2f2b268d43df2b80db942c. I think kses thinks its invalid attributes.

Issue #45, opened 2014-11-11
Show (live) preview when adding item


stompweb:
It would be good if there was a preview (preferably a live preview) of the shortcode output that was being entered (in the instance that there was maybe 10-15 shortcodes). The would be good on sites with a lot of shortcodes so the user knew what they were producing. It also means the user doesn't have to add it and then view it in TinyMCE to see the output.

![screenshot 2014-11-11 15 32 02](https://cloud.githubusercontent.com/assets/1017511/4995313/f8fed918-69b7-11e4-910d-3f644ddcc6a8.png)

mattheu:
+1 Thanks!

danielbachhuber:
:+1:

danielbachhuber:
:ship: :ship: :ship:

PR #61, opened 2014-11-24
Implement Live Preview in Shortcake UI

zebulonj:
This branch adds a live preview to the Shortcake UI, per #45. The preview is currently un-styled, appearing at the top of the UI (under the 'Edit' title).

screen shot 2014-11-24 at 7 19 04 am

This implementation is not yet tested against more advanced shortcodes or Shortcake inputs other than &lt;input> and &lt;textarea>. I'm opening this pull request primarily to facilitate review of current progress and discussion of next steps.

Note that this feature is impacted by the missing attributes referenced in #60.


zebulonj:
@danielbachhuber I think this is ready for initial review. I think it could use some UX polish and I'd like to add some automated unit testing for the JavaScript I've added, but wanted to give you the change to validate that this is what the team had in mind.

danielbachhuber:
UX feedback:
- [x] When the preview hasn't yet loaded, there should be a placeholder. Stealing what core did for TinyMCE would be fine.
- [x] Preview should be styled as it would be in the visual editor (e.g. load up the editor stylesheet).
- [x] There should be some visual distinction between the preview area and the fields area. I'm not sure "Edit" right above the preview makes the most sense.

noppanit:
@zebulonj Could you please add the test you mentioned? You can choose any tool you like (e,g qUnit or Jasmine).

js/shortcode-ui.js
```diff
@@ -231,6 +235,89 @@ var Shortcode_UI;
...
+			var data;
+			var shortcode = this.model.clone();
+			
+			// Use wp.template style delimiters. Would prefer not to duplicate this option data, 
+			// but can't use wp.template directly on an actual template string.
+			var options = {
+				evaluate:    /<#([\s\S]+?)#>/g,
+				interpolate: /\{\{\{([\s\S]+?)\}\}\}/g,
+				escape:      /\{\{([^\}]+?)\}\}(?!\})/g,
+				variable:    'data'
```
noppanit: @zebulonj What do you think would be a better way of doing this?

danielbachhuber:
> You can choose any tool you like (e,g qUnit or Jasmine).

We should use Jasmine for consistency with what we're already using.

zebulonj:
@noppanit Are you referring to the approach I used for fetching/generating the template, or to my notation about duplication of the `_.template(...)` options from `wp.template(...)`?

noppanit:
@zebulonj the duplication of wp.template. I was just wondering if there's a better approach.

zebulonj:
@danielbachhuber Agreed on the UX enhancements... delineation of the preview, styling consistent with editor, and the loading placeholder. Just wanted to get feedback on direction before I went too far down a path.

zebulonj:
@noppanit Thanks for clarifying.

There are definitely other possible approaches. Due to trade-offs, I think it's debatable which are better.

`wp.template` assumes that the template text has been rendered into a targetable &lt;script> block in the page source (as most of the templates used in the Shortcake UI are), and pulls the text to be compiled accordingly. I explored rendering the shortcode markup into the page source for this purpose, using something along the lines of a server-side call to `do_shortcode(...)`, but see some challenges and some shortcomings in that approach:
1. All available shortcodes would have to be templatized in this way at the time the page source is rendered, so that any one of them can be used within the Shortcode UI.  Not necessarily a bad thing, but requires additional server-side logic to inject Underscore.js template tokens into the text before it is rendered into the page. Also, requires additional client-side logic to identify and target the template for the selected shortcode.
2. More importantly, when previewed in TinyMCE, Shortcake pulls the shortcode markup asynchronously. Re-using this functionality to generate the client-side template for the live preview serves the need of consistency between the two previews and avoids duplicating code used in [`handle_ajax_do_shortcode(...)`](https://github.com/fusioneng/Shortcake/blob/master/inc/class-shortcode-ui.php#L182-L201). There is the trade-off of not being able to use `wp.template(...)`. However, [`wp.template(...)`](https://github.com/WordPress/WordPress/blob/master/wp-includes/js/wp-util.js#L17-L36) is mostly just a convenience method that wraps `_.template(...)` and fetches the template text from the DOM using a selector. Really, in my code, the only thing that is duplicated is the options that are passed to `_.template(...)`.

When balancing the trade-offs, I erred in favor of being consistent about how the shortcode is rendered for preview: using [`handle_ajax_do_shortcode(...)`](https://github.com/fusioneng/Shortcake/blob/master/inc/class-shortcode-ui.php#L182-L201).

If there's something I've overlooked, or if you think the trade-offs favor a different approach, I want to hear about it. This kind of dialog leads to well thought-out code and I work best when challenged by my peers.

noppanit:
Thanks @zebulonj this makes a lot more sense.

zebulonj:
@danielbachhuber @noppanit The test setup is going to be dependent on the restructuring method, so I've done the tool configuration and added a sample spec on the restructuring branch (#64).

danielbachhuber:
@zebulonj Let me know when this is ready for a final review

zebulonj:
@danielbachhuber I'm confronting challenges when it comes to adding the editor styles to the preview in the media UI. If editor stylesheets are to be applied to the preview content, it seems it should be sandboxed in its own iFrame (the editor stylesheets shouldn't be added to main page). [The code you pointed me to](https://github.com/fusioneng/Shortcake/blob/master/js/shortcode-ui.js#L536) would accomplish this, but re-use doesn't appear to be an option outside the context of the TinyMCE editor (such as in the media UI). It relies heavily of TinyMCE infrastructure. I'm sure I could find an answer, given more time, but for now I suggest that we forgo applying the editor styles to the preview.

danielbachhuber:
> If editor stylesheets are to be applied to the preview content, it seems it should be sandboxed in its own iFrame (the editor stylesheets shouldn't be added to main page).

This makes sense.

> The code you pointed me to would accomplish this, but re-use doesn't appear to be an option outside the context of the TinyMCE editor (such as in the media UI).

As an imperfect, but mostly functional workaround, why not sniff out the custom stylesheets registered to the `$editor_styles` global and return those in the rendered iframe?

> I'm sure I could find an answer, given more time, but for now I suggest that we forgo applying the editor styles to the preview.

Having properly styled shortcodes is a requirement for providing a preview.

zebulonj:
> As an imperfect, but mostly functional workaround, why not sniff out the custom stylesheets registered to the $editor_styles global and return those in the rendered iframe?

Sorry. I should have been more specific. The challenge isn't identifying the CSS to be added to the iFrame. The challenge is in manipulating the iFrame content cross-browser. There's functionality for this in TinyMCE, but because of the way it is architected, it won't work outside the editor. I'm aware of one third-party library that purports to support his (http://travistidwell.com/seamless.js/), but I have never used it and can't vouch for it. In any event, this update is just FYI. I'm trying to work out a solution.

zebulonj:
I've isolated the preview content in an &lt;iframe>, applied the editor styles in that sandbox, added a loading indicator to match the TinyMCE view, and given the UX a light polish to separate the preview from the form.

screen shot 2014-11-30 at 11 33 32 pm

There is some additional work to be done:
- [x] The &lt;iframe> manipulation utility needs to be browser tested.

I have some recommendations for enhancement to the UX. The existing preview and form UX works will for a simple (vertically small when rendered) shortcode, in which case the preview and form are both visible together in a normal sized browser window. For shortcodes that generate larger rendered content, this will not be the case:  the preview and form will not be visible at the same time and the form will not show initially when the UI opens. I can envision two options that would provide a good user experience in this case:
1. Separating the form and preview to tabs, similar to the Github "write" and "preview" tabs (issue comments, etc).
2. Make the preview area collapsible. Possibly sliding up and out of the frame, leaving a "preview" tab hanging down so that it can be expanded. Possibly defaulting to collapsed so that the form is immediately visible when the UI opens.

css/sass/shortcode-ui-editor-styles.scss
```diff
@@ -8,3 +8,41 @@
 		}
 	}
 }
+
+.loading-placeholder {
```
danielbachhuber: Should we namespace this?

inc/templates/edit-form.tpl.php
```diff
@@ -1,14 +1,16 @@
 <script type="text/html" id="tmpl-shortcode-default-edit-form">
 
-	<form class="edit-shortcode-form">
-
-		<h2 class="edit-shortcode-form-title">Edit {{ data.label }}</h2>
+	<div class="preview-shortcode">
+		<h2 class="preview-shortcode-title">Preview</h2>
```
danielbachhuber: We should make these strings translatable.

css/sass/shortcode-ui-editor-styles.scss
zebulonj: > Should we namespace this?

Yes. Because the CSS is not only loaded in the preview &lt;iframe> (sandbox). It is also loaded in the editor where it could potentially conflict with other styles.

danielbachhuber:
I don't think this quite works as I'd expect it to. When I update a shortcode's attributes, the preview should make an ajax request to WordPress to get the rendered shortcode.

Many shortcodes have functional logic in their PHP callbacks that would be redundant to re-implement in JavaScript. For instance, our `[fusion-content id=24243]` shortcode gets a video of ID 24243 from the database, and then renders an embed code based on whether it's an Ooyala or YouTube video.

Sorry if I missed this before.

I agree the preview UI could use some work.

zebulonj:
> When I update a shortcode's attributes, the preview should make an ajax request to WordPress to get the rendered shortcode.

@danielbachhuber That's an easy fix. I initially built this on top of an AJAX request for the template (rather than baking the template into a &lt;script> block the way the other client-side templates are delivered) for exactly that reason:  that the content of the rendered shortcode may change based on the post associated with or the attributed provided. I think the only thing we're currently doing wrong is that once initially fetched via ajax, the template is cached client-side to make the "live" updates work. In light of your point (a good one), I don't think it's appropriate to have the preview update as the user types. Any delay is server-side response will result in back-pressure and potential state conflicts as the user types. All the more reason to switch to a preview on a separate tab, a-la the Github interface.

danielbachhuber:
Ok, sounds good.

inc/templates/edit-form.tpl.php
zebulonj: > We should make these strings translatable.

Done.

zebulonj:
Preview is now removed to its own tab, which refreshes with a call to the server each time it is opened.

screen shot 2014-12-01 at 3 24 44 pm
screen shot 2014-12-01 at 3 24 52 pm

zebulonj:
@danielbachhuber This is browser tested back it IE9 and appears to be solid. I can spend some additional time testing against more complex shortcodes (more complex than the shortcake demo linked from the README) if you think that's called for.

danielbachhuber:
Nah, IE9 is good enough

zebulonj:
@danielbachhuber This is ready for review.

Note: I've added documentation to the new classes and methods that uses [JSDoc](http://usejsdoc.org) syntax for automatic documentation generation.

css/sass/shortcode-ui.scss
```diff
@@ -114,3 +151,8 @@
 	}
 
 }
+
+.preview-shortcode {
+	padding: 0 20px;
+	border-bottom: 1px solid #666;
+}
```
danielbachhuber, line 128: This should be namespaced too, no?

js/shortcode-ui.js
```diff
@@ -265,10 +500,23 @@ var Shortcode_UI;
...
+			var shortcode = this.controller.props.get( 'currentShortcode' );
+			var view = new sui.views.TabbedView({
+				tabs: {
+					edit: {
+						label: "Edit",
+						content: new sui.views.EditShortcodeForm({ model: shortcode })
+					},
+
+					preview: {
+						label: "Preview",
```
danielbachhuber: These labels should be translatable.

js/shortcode-ui.js
```diff
@@ -231,8 +386,88 @@ var Shortcode_UI;
...
+
+			_.each( tinymce.editors, function( editor ) {
+				_.each( editor.dom.$( 'link[rel="stylesheet"]', editor.getDoc().head ), function( link ) {
+					var href;
+					( href = link.href ) && ( styles[href] = true );
+				});
+			});
+
+			styles = _.map( _.keys( styles ), function( href ) {
+				return '<link rel="stylesheet" type="text/css" href="' + href + '">';
```
danielbachhuber: VIP is going to ask that this HTML be built using jQuery, which will appropriately escape `href`

danielbachhuber:
Getting close! My shortcode returns this response for the preview, which doesn't work:

```
<div class="fusion-video-embed"><div id='DYejXXX_5c4' data-video-type='youtube'></div></div><script src='https://www.youtube.com/iframe_api'></script>
<script>
var ytplayer0;
function onYouTubeIframeAPIReady() {
ytplayer0 = new YT.Player("DYejXXX_5c4", {
videoId: "DYejXXX_5c4",
events: {
'onReady': onPlayerReady
},
playerVars: {
rel: 0
}
});
}
function onPlayerReady(event) {
}
</script>
```

I'm not a huge fan of how the tabs look, but I think it's good enough for now.

js/shortcode-ui.js
zebulonj: Resolved.

css/sass/shortcode-ui.scss
zebulonj, line 128: Yes. This was replaced with styles at line 74, but not cleaned up. Thanks.

js/shortcode-ui.js
```diff
@@ -231,8 +386,88 @@ var Shortcode_UI;
...
+
+			_.each( tinymce.editors, function( editor ) {
+				_.each( editor.dom.$( 'link[rel="stylesheet"]', editor.getDoc().head ), function( link ) {
+					var href;
+					( href = link.href ) && ( styles[href] = true );	// Poor man's de-duping.
+				});
+			});
+
+			styles = _.map( _.keys( styles ), function( href ) {
+				return $( '<link rel="stylesheet" type="text/css">' ).attr( 'href', href );
```
zebulonj: > VIP is going to ask that this HTML be built using jQuery, which will appropriately escape `href`.

Is this what you had in mind?

danielbachhuber: Egg-axctly.

zebulonj:
@danielbachhuber Regarding your shortcode that doesn't work... can you point me to a plugin or plugins so that I can work against the same environment you're seeing the issue with? Also, is it failing in both the embedded MCE view and the preview, or is it just failing in the preview?

danielbachhuber:
> can you point me to a plugin or plugins so that I can work against the same environment you're seeing the issue with?

Did you create a shortcode that just returns what I provided as the response? That should be sufficient.

> Also, is it failing in both the embedded MCE view and the preview, or is it just failing in the preview?

Just the preview.

zebulonj:
> Did you create a shortcode that just returns what I provided as the response? That should be sufficient.

I was headed down that path when I realized that the JavaScript in the shortcode you referenced depends on external JavaScript. My suspicion is that this dependency is at the heart of the issue (I've put some plain JavaScript in a shortcode without problem), and that it is related to the sandboxing. If it is not convenient to share the source of that shortcode, I can create my own shortcode that references external JavaScript.

zebulonj:
@danielbachhuber If the YouTube shortcode is working in the MCE View, but not the preview, I suspect that the plugin is doing something to register the external JavaScript for inclusion in the MCE View. If I can identify what mechanism it is using, I can reproduce that in the preview sandbox. I've tested a shortcode of my own making, which relies on external JavaScript, and it fails in both the MCE View and the preview because of the missing dependency.

danielbachhuber:
Here's the TinyMCE view preview in the editor:

image

The DOM looks like this:

image

The snippet I copy and pasted above was a direct copy and paste from the admin ajax response:

image

(The ajax response for the modal preview looks the exact same).

danielbachhuber:
I would test to see if adding a DOCTYPE to the iframe fixes it — it might be that the YouTube script isn't firing because it doesn't think the document is ready or something.

Or, wrap in `setTimeout()` to see if something needs to get ready.

It would be nice if the existing iframe generation code was abstracted out for reusability: https://github.com/fusioneng/Shortcake/blob/master/js/shortcode-ui.js#L550

zebulonj:
The issue reported by @danielbachhuber above--where a YouTube embed didn't work in the preview--was due to the way iframe content was updated. When markup that includes scripts is inserted into a ready DOM, the execution order of those scripts is unpredictable (at best). The issue was resolved by writing a complete HTML document to the iframe.

Recommend creating a new issue to complete abstraction (to the extent possible) of the iframe rendering code. I've written the iframe code used in the preview (in `sui.dom.IFrame`), with a mind toward future re-use to replace portions of the code in the MCE View constructor. Caution should be used, because certain portions of the MCE View constructor use code specific to the TinyMCE environment.

danielbachhuber:
:+1: confirmed the fix.

@mattheu @noppanit can you guys give this a final review?

mattheu:
Played around with this and its working OK I have a few comments though.

**Implementation**

I've been working on a shortcode that requires external JS to work. Getting it to work in the editor was difficult and a little hacky. 

My concern here is that I would now need a 3rd workaround to get it displaying correctly in the preview.

But - I do think the iFrame is a good idea and it might be a route to solving both problems - by using the same iFrame in the tinyMCE editor.

**UX**

The tabs need a little polish to make them blend in better with the WordPress UI.

My initial reaction was they should be the same as the built in media modal tabs - see screenshot. 

image

However i'm not convinced by the tabs. I actually liked having the preview on the same view - it seems that if you have to click 'preview' you may as well click 'update/insert' and preview inside the editor. I'm sure we can listen for changes to the shortcode model and re-render the preview. We may need to watch for changes in the fields as I think it currently updates on insert...

zebulonj:
@mattheu Thanks for reviewing. Regarding placing the preview on a separate tab...

Early in development on this branch, I did embed the preview in the same view as the form. As you suggest above, I listened for changes on the shortcode model and re-rendered the preview accordingly. I moved it to its own tab for two reasons.
1. Rendered shortcodes could vary significantly in size. If the rendered content is tall, it will push the form out of view. If the two can't be viewed simultaneously without scrolling, it might as well be moved to its own tab and it seems like a poor UX to not have the form immediately visible when the UI is opened.
2. Rendering of the preview must be done server-side, to account for shortcodes that pull in additional content based upon attributes set by the user. At least in development so far, that takes ~500ms (maybe more). If the client requests a fresh rendering each time the model changes (each time a key is pressed), things are going to get messy with lots of requests to the server. That could be managed client-side by debouncing or something to that effect, if it was the only reason for moving the preview to its own tab (and, if so, I'd recommend that be targeted for the next iteration).

danielbachhuber:
Let's clean up the tabs on this branch, and open a new issue to discuss consolidating preview and editing.

zebulonj:
Tabs updated. 

screen shot 2014-12-03 at 12 10 50 pm
screen shot 2014-12-03 at 12 11 04 pm

It may be worth considering--if we stick with a tabbed interface--incorporating the tabs in the the media router interface. It might also be nice to move the "return to list view" link into the tab bar. I think the current implementation is sufficient to get this enhancement out the door.

mattheu:
This looks loads better - thanks!

js/shortcode-ui.js
```diff
@@ -11,6 +11,102 @@ var Shortcode_UI;
...
+			 * Creates an <iframe> appended to the `$parent` node. Manages <iframe> content updates and state.
+			 *
+			 * @method create
+			 * @static
+			 * @param $parent jQuery object to which <iframe> should be appended.
+			 * @param [options.body] HTML content for insertion into the &lt;iframe> body.
+			 * @param [options.head] Markup for insertion into the &lt;iframe> head.
+			 */
+			create: function( $parent, options ) {
+				var iframe	= document.createElement( 'iframe' );
```
noppanit, line 35: @danielbachhuber Is VIP ok with `document.createElement`?

js/shortcode-ui.js
```diff
@@ -231,8 +422,115 @@ var Shortcode_UI;
...
+
+			// Fetch shortcode markup using template tokens.
+			data = {
+				action: 'do_shortcode',
+				post_id: $('#post_ID').val(),
+				shortcode: shortcode.formatShortcode(),
+				nonce: shortcodeUIData.previewNonce
+			};
+
+			$.post( ajaxurl, data, callback );
```
noppanit, line 506: @zebulonj This is just a question. What do you think of abstracting this out so we have like a service layer so we can test it piece by piece separately? Then we can just inject that into this method so we can mock that later in the test?

js/shortcode-ui.js
```diff
@@ -176,16 +272,112 @@ var Shortcode_UI;
...
+
+	/**
+	 * Single edit shortcode content view.
+	 */
+	sui.views.EditShortcodeForm = wp.Backbone.View.extend({
 		template: wp.template('shortcode-default-edit-form'),
 
 		initialize: function() {
-
 			var t = this;
```
noppanit, line 378: @zebulonj for `var t = this;` did you mean `self` or it has special meaning?

zebulonj, line 378: @noppanit I prefer the `var self = this;` convention. The quoted line is prior existing code.

js/shortcode-ui.js
zebulonj, line 506: @noppanit Good suggestion.

zebulonj, line 506: Created new issue: #68.

js/shortcode-ui.js
danielbachhuber, line 35: Yep, this is fine.

danielbachhuber:
:ship: :ship: :ship:

PR #63, opened 2014-11-24
Abstract field registration

mattheu:
Continuing the work from https://github.com/fusioneng/Shortcake/pull/44

Adds `Shortcode_UI_Fields` class to manage field types. Each field type requires a template and a view model to be specified. The field controller then outputs field data that can be used by the JS.

Added a post select field.

The main aim is around better abstraction allowing for developers to create their own fields. I think that providing support for fieldmanager fields as a shortcake extension is the ultimate test of the flexibility.


dev.php
```diff
@@ -28,11 +35,11 @@
 	 * and an array or args.
 	 */
 	shortcode_ui_register_for_shortcode(
-		'pullquote',
+		'blockquote',
```
danielbachhuber: Merge conflict

dev.php
```diff
@@ -28,11 +35,11 @@
...
 		array(
 
 			// Display label. String. Required.
-			'label' => 'Pullquote',
+			'label' => 'Dev Blockquote',
```
danielbachhuber: Merge conflict

inc/fields/class-fieldmanager-fields.php
```diff
@@ -0,0 +1,165 @@
...
+
+	static $instance;
+
+	private $fields = array(
+		'Fieldmanager_Textfield' => array(
+			'template' => 'shortcode-ui-field-Fieldmanager_Textfield',
+		),
+		'Fieldmanager_TextArea' => array(
+			'template' => 'shortcode-ui-field-Fieldmanager_TextArea',
+		),
```
danielbachhuber: Why is this one declared twice?

mattheu: Just the way fieldmanager named the fields - `Fieldmanager_Textfield` != `Fieldmanager_TextArea`

mattheu:
Fixed conflicts

Also removed fieldmanager fields. I think fieldmanager fields should be optional so that it is not a requirement for using shortcake - either as a separate plugin, or at least check for fieldmanager before enabling.

I think this can be added as a separate PR anyway.

dev.php
```diff
@@ -6,24 +6,29 @@
 	 * Register your shortcode as you would normally.
 	 * This is a simple example for a pullquote with a citation.
 	 */
-	add_shortcode( 'pullquote', function( $attr, $content = '' ) {
+	add_shortcode( 'shortcake_dev', function( $attr, $content = '' ) {
```
danielbachhuber, line 9: Did you mean to change this?

mattheu, line 9: This is just a helper file for development - It shouldn't normally be included - it just includes a single shortcode with all the different fields. I think its useful for anyone wanting to get started with shortcake.

Problem is - we have a 'pullquote' shortcode on Fusion - and they conflict.

danielbachhuber, line 9: That's fine. Can we have it in a separate PR though?

inc/fields/class-fields.php
```diff
@@ -0,0 +1,214 @@
+<?php
+
+class Shortcode_UI_Fields {
```
danielbachhuber: If we're going to call this class `Shortcode_UI_Fields`, then the file name should be `class-shortcode-ui-fields.php` and probably makes most sense in `/inc`

js/lib/select2/bower.json
```diff
@@ -0,0 +1,8 @@
+{
```
danielbachhuber: Let's drop the Select2 files we aren't using. No sense in including a bunch of stuff that's not in use.

shortcode-ui.php
```diff
@@ -20,9 +20,14 @@
  */
 
 require_once dirname( __FILE__ ) . '/inc/class-shortcode-ui.php';
+require_once dirname( __FILE__ ) . '/inc/fields/class-fields.php';
+require_once dirname( __FILE__ ) . '/dev.php';
```
danielbachhuber: Probably shouldn't load this by default.

danielbachhuber:
This seems more or less fine. I'm still very interested to see Fieldmanager integration, though.

mattheu:
I've updated this to fix the issues you flagged - but also to move the post select field to a separate file.

danielbachhuber:
> but also to move the post select field to a separate file.

How about we just keep it in the Fusion theme for now? This way, we can use the Select2 library we've already internalized

mattheu:
Makes sense. I'll remove.

mattheu:
Updated to remove post select. Also merged master.

danielbachhuber:
:+1:

Issue #55, opened 2014-11-18
Test coverage for the JavaScripts


danielbachhuber:
Coming out of #54, we should have test coverage for the important parts.

noppanit:
@danielbachhuber I think it's a good idea to restructure our js as well. It's 800 lines of code already. Should it be the task of this issue or it can be a separate issue?

danielbachhuber:
> It's 800 lines of code already. Should it be the task of this issue or it can be a separate issue?

It can be the task of this issue.

zebulonj:
@danielbachhuber @noppanit Would love to discuss what the restructuring of the JavaScript might look like. I will follow-up later with recommendations (short-term and long-term).

noppanit:
@zebulonj sounds good look forward to your recommendations.

zebulonj:
@danielbachhuber @noppanit Hoping to get some initial feedback on the following recommendation before you shut down for the holiday.
1. I recommend using [Require.js](http://requirejs.org/) to support modularizing the code base and to manage dependencies. The Require.js [optimizer](http://requirejs.org/docs/optimization.html) would be used to package the source into a single file, as part of the build process (along with compiling the Sass).
2. I recommend organizing components (and files) around feature oriented modules, rather than a top level organization around role (model, view, controller, template). As a code-base grows, having models, views, controllers, etc., associated with a feature (for example, 'preview') grouped together, rather than spread across 'model', 'view', and 'controller' folders makes it easier for developers to work in the code and makes it easier to reuse individual features across projects.

I will setup a skeleton of what this might look like, but wanted to get some notes up here quickly for your input.

noppanit:
Thank you @zebulonj for the recommendation. What do you think about http://browserify.org/ vs Require.js?

Just to clarify my comment. I hit some limitation of require.js sometime when I needed to use plugin that's not require.js compatible. Also, browserify syntax is like Node.js

However, I want to here your thoughts as well.

zebulonj:
@noppanit I don't have first-hand experience with Browserify, but based on what I've read I'd be interested in trying it. 

[...]

PR #64, opened 2014-11-26
Restructure JS and Add Test Coverage

zebulonj:
IN PROGRESS.

See #55.

### DIFF
--- a/js/src/mce/mce.view.constructor.js
+++ b/js/src/mce/mce.view.constructor.js
@@ -1,0 +1,294 @@
+define( ['jquery'], function ( $ ) {
+	'use strict';
+
+	// @was sui.utils.shortcodeViewConstructor
+	var MCEViewConstructor = {
+
+		View: {
+
+			shortcodeHTML: false,
+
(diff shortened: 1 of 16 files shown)


zebulonj:
@danielbachhuber @noppanit A functional skeleton of what restructuring with Require.js could look like is visible under [`js/src`](https://github.com/fusioneng/Shortcake/tree/feature/restructure/js/src). I've also added the Grunt build configuration for Require.js and have configured JSHint. Feedback welcome. Still happy to explore using Browserify.

js/src/mce/mce.view.constructor.js
```diff
@@ -0,0 +1,296 @@
+define( ['jquery'], function ( $ ) {
```
noppanit, line 1: @zebulonj by the way I quite like the pattern of using 

```
define(function(require) {
  var $ = require('jquery');
}
```

It can get messy when a lot of modules have to be imported.

danielbachhuber:
We'll be using Browserify when we move forward with restructuring.

PR #72, opened 2014-12-10
Pass shortcode to shortcode preview actions

danielbachhuber:
This will make it easier for callbacks to target to specific shortcodes


danielbachhuber:
@mattheu thoughts?

mattheu:
Looks good

PR #78, opened 2014-12-11
Fix fieldmanager media field

mattheu:
Fieldmanager media field is broken.

The cause is that the JS programatically updates the hidden input field, and the normal events that trigger the updating of the model don't fire.

So we need to do this manually - when the frame is closed. This is a little hacky in that we have to wait for the frame to be created before attaching the event - hence the setTimeout - but it works!

Review & Merge cc @danielbachhuber

### DIFF
--- a/inc/fields/class-shortcode-ui-fields-fieldmanager.php
+++ b/inc/fields/class-shortcode-ui-fields-fieldmanager.php
@@ -144,7 +144,8 @@ class Shortcode_UI_Fields_Fieldmanager {
 			}
 
 			/**
-			 * Update the value when media iframe is closed.
+			 * Update the value when media iframe is closed,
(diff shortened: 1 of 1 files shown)


inc/fields/class-shortcode-ui-fields-fieldmanager.php
```diff
@@ -143,6 +143,20 @@ public function action_print_media_templates() {
 
 			}
 
+			/**
+			 * Update the value when media iframe is closed.
```
danielbachhuber: Should we include the comment about why we're using `setTimeout()`?

danielbachhuber:
How was this working for you before?

danielbachhuber:
Still seems kinda broken (e.g. doesn't work with preview)

PR #91, opened 2014-12-22
Build strings with jQuery to mitigate XSS

### LINKED ISSUE #77: DOM should be constructed with jQuery instead of strings
The following DOM should be constructed with jQuery, instead of strings, to avoid XSS:

```
var previewHTML = localization.strings.uploadedFile + ':</br>';
previewHTML += '<a href="#">' + response.data.html + '</a></br>';
previewHTML += '<a class="fm-media-remove fm-delete" href="#">' + localization.strings.removeFile + '</a>';
```

danielbachhuber:
Fixes #77


danielbachhuber:
@mattheu #reviewmerge

mattheu:
Looks good

Issue #90, opened 2014-12-19
Conflict with Advanced Custom Fields Plugin


tmconnect:
If using both plugins together I got two javascript errors using the ACF field settings

[Error] TypeError: undefined is not an object (evaluating 'wp.media.view.settings.post.id')
    (anonyme Funktion) (load-scripts.php, line 157)
    global code (load-scripts.php, line 157)
[Error] TypeError: undefined is not an object (evaluating 'wp.mce.views')
    (anonyme Funktion) (shortcode-ui.js, line 1074)
    forEach ([native code], line 0)
    forEach (load-scripts.php, line 54)
    (anonyme Funktion) (load-scripts.php, line 55)
    (anonyme Funktion) (shortcode-ui.js, line 1070)
    j (load-scripts.php, line 2)
    fireWith (load-scripts.php, line 2)
    ready (load-scripts.php, line 2)
    J (load-scripts.php, line 2)

coderaaron:
Here is the quick fix I use:

change 

```
        sui.shortcodes.each( function( shortcode ) {
            // Register the mce view for each shortcode.
            // Note - clone the constructor.
            wp.mce.views.register(
                shortcode.get('shortcode_tag'),
                $.extend( true, {}, sui.utils.shortcodeViewConstructor )
            );
        } );
```

to

```
        sui.shortcodes.each( function( shortcode ) {
            if( wp.mce ) {
                // Register the mce view for each shortcode.
                // Note - clone the constructor.
                wp.mce.views.register(
                    shortcode.get('shortcode_tag'),
                    $.extend( true, {}, sui.utils.shortcodeViewConstructor )
                );
            }
        } );
```

starting at line 1070 in shortcode-ui.js

Be warned I haven't done extensive testing with this "hack", but it seems to work for what I've been doing.

danielbachhuber:
Fixed in #118

PR #95, opened 2014-12-26
Make Shortcake work with ACF plugin

tmconnect:
I changed this in my installation to solve issue #90 to make it work with the Advanced Custom Field plugin.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -72,7 +72,9 @@ class Shortcode_UI {
 
 	public function action_admin_enqueue_scripts( $hook ) {
 
-		if ( in_array( $hook, array( 'post.php', 'post-new.php' ) ) ) {
+		global $current_screen;
+
+		if ( in_array( $hook, array( 'post.php', 'post-new.php' ) ) && $current_screen->post_type <> 'acf-field-group') {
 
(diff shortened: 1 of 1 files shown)


inc/class-shortcode-ui.php
```diff
@@ -72,7 +72,9 @@ public function get_shortcode( $shortcode_tag ) {
 
 	public function action_admin_enqueue_scripts( $hook ) {
 
-		if ( in_array( $hook, array( 'post.php', 'post-new.php' ) ) ) {
+		global $current_screen;
+
+		if ( in_array( $hook, array( 'post.php', 'post-new.php' ) ) && $current_screen->post_type <> 'acf-field-group') {
```
danielbachhuber, line 77: Better to use `get_current_screen()` than the `$current_screen` global. When you read from global variables, you run the chance of accidentally stomping on them and causing problems later.

danielbachhuber:
I'm not keen on hardcoding a workaround for ACF. Could you describe why the JS error is occurring (screenshots are helpful) so I have a better sense of the root of the problem?

danielbachhuber:
@tmconnect now that #118 is merged, can you test master branch and confirm the fix?

tmconnect:
@danielbachhuber I have tested with ACF 5.1.8 and now it's working and there are no conflicts.

danielbachhuber:
Thanks for the confirmation, @tmconnect

PR #103, opened 2015-01-09
Use the instance to get the shortcode model.

ellatrix:


danielbachhuber:
Thanks, @avryl !

@mattheu can you test this?

danielbachhuber:
@avryl If you remove `setHTML()` in 527fa9213d2b16b88da77382cbea8259c939d6c7, how are we supposed to accommodate for shortcodes that load scripts and other nefarious HTML?

ellatrix:
Better to use `setIframes` then, but no need to copy those methods for core I think.

mattheu:
Thanks for this Avryl

One issue is that I’m getting the following error: 

Uncaught TypeError: undefined is not a function in media-editor.js line 839

This seems to be caused by this line: 

this.trigger( 'insert', shortcode );

If I revert this line only it all works OK.

danielbachhuber:
Continued in #143

PR #108, opened 2015-01-14
Use named function to switch tabs when preview

jitendraharpalani58:
See: https://github.com/fusioneng/fusion-theme/issues/888


danielbachhuber:
:+1:

PR #112, opened 2015-01-20
remove TinyMCE filter for stylesheet; use add_editor_style instead

crstauf:
fix Shortcake overwriting other TinyMCE stylesheets


danielbachhuber:
If we're no longer using it, can you remove the `modify_tiny_mce_4()` method too?

crstauf:
sorry i failed to do that the first time; done now.

danielbachhuber:
Looks good :+1: Thanks!

PR #117, opened 2015-01-29
Remove conflict with ACF Pro

coderaaron:
From what I can tell Advanced Custom Fields doesn't create the default (or any) TinyMCE instance that `wp.mce.views.register` attempts to use.

This is almost too simple a fix, but it seems to work fine on all my installs.  I pasted it in https://github.com/fusioneng/Shortcake/issues/90 last month, but I thought I should create a pull request to at least try to give back.

### DIFF
--- a/js/shortcode-ui.js
+++ b/js/shortcode-ui.js
@@ -1075,14 +1075,14 @@ var Shortcode_UI;
 		sui.shortcodes = new sui.collections.Shortcodes( shortcodeUIData.shortcodes )
 
 		sui.shortcodes.each( function( shortcode ) {
-
-			// Register the mce view for each shortcode.
-			// Note - clone the constructor.
-			wp.mce.views.register(
-				shortcode.get('shortcode_tag'),
-				$.extend( true, {}, sui.utils.shortcodeViewConstructor )
-			);
-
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@coderaaron now that #118 is merged, can you test master and confirm the fix?

danielbachhuber:
Confirmed in #95

js/shortcode-ui.js
```diff
@@ -1075,14 +1075,14 @@ var Shortcode_UI;
...
 		sui.shortcodes.each( function( shortcode ) {
-
-			// Register the mce view for each shortcode.
-			// Note - clone the constructor.
-			wp.mce.views.register(
-				shortcode.get('shortcode_tag'),
-				$.extend( true, {}, sui.utils.shortcodeViewConstructor )
-			);
-
+			if( wp.mce ) {
```
danielbachhuber: Can we actually check `wp.mce.views` here, in the case that `mce` is set and `views` is not?

coderaaron:
OK, should be good to go.  Thanks for your help on this.  Let me know if there is anything else I can do.

danielbachhuber:
Thanks again @coderaaron

PR #143, opened 2015-02-08
JS cleanup

mattheu:
Picking this up - https://github.com/fusioneng/Shortcake/pull/103 - there were some great improvements but it wasn't actually working correctly.
- Reverted `this.trigger( 'insert', shortcode );` addition - I couldn't work out what this should be.
- Updated ajax to use wp.ajax and to handle errors.
- Added styles for errors
- Fixed styles not loading.

Thanks for your work on this @iseulde - sorry it took me so long to dig into it a bit more.


danielbachhuber:
:+1:

Issue #106, opened 2015-01-12
Access Media


Danny159:
Have the ability to select a madia file :)

danielbachhuber:
@Danny159 Not sure I follow. Could you explain in greater detail?

Danny159:
Sorrry @danielbachhuber, Have a button next to a text box saying Add Media, where you can then select an image from the Media Library and it puts the URL into the textbox :)

Hope you now understand :)

danielbachhuber:
What if we had a rich textarea that behaved like the normal WordPress editor?

Danny159:
Hmm, Yes I guess you are right... :)

soderlind:
@danielbachhuber not sure if I agree. If I only need to add media, why should I have/see the wp editor?

I'd prefer a field like this (screenshot from the [Admin Page Framework](https://github.com/michaeluno/admin-page-framework)):

![media-file](https://cloud.githubusercontent.com/assets/1649452/6011468/7c30ad0a-ab3c-11e4-8650-a9cfee7d1446.png)

milgrom:
I agree that an add media button would be a great addition to keep things simple for non technical users.  And the rich textarea editor would also be a great addition for situations where it's required.

yumyo:
+++1

danielbachhuber:
Added in #146 as `type=attachment`

PR #146, opened 2015-02-11
Image/Attachment field

mattheu:
I've had a few requests for an image field that doesn't require the whole of fieldmanager for those who don't use it.

@todo Should handle restricting library by file type

See #106


css/sass/shortcode-ui.scss
```diff
@@ -128,3 +132,68 @@
 
 }
 
+
+.edit-shortcode-form .shortcake-attachment-preview {
+
+	width: 150px !important;
+	height: 150px !important;
```
danielbachhuber: Why do we need these `!important`?

dev.php
```diff
@@ -44,8 +49,8 @@
 			// Display label. String. Required.
 			'label' => 'Shortcake Dev',
 
-			// Icon/image for shortcode. Optional. src or dashicons-$icon. Defaults to carrot.
-			'listItemImage' => 'dashicons-editor-quote',
+			// Icon/attachment for shortcode. Optional. src or dashicons-$icon. Defaults to carrot.
+			'listItemAttachment' => 'dashicons-editor-quote',
```
danielbachhuber: Why is this example changing?

inc/fields/class-field-attachment.php
```diff
@@ -0,0 +1,77 @@
...
+	 */
+	public function action_print_media_templates() {
+
+		?>
+
+		<script type="text/html" id="tmpl-fusion-shortcake-field-attachment">
+			<div class="field-block">
+				<label for="{{ data.attr }}">{{ data.label }}</label>
+				<div class="shortcake-attachment-preview attachment-preview attachment">
+					<button id="{{ data.attr }}" class="button button-small add">Select Attachment</button>
```
danielbachhuber: String should be translatable.

inc/fields/class-field-attachment.php
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Initialize Post Fields.
+	 *
+	 * Find all registered shortcode attributes that use the post field and
+	 * store an array with all their args for future reference.
+	 *
+	 * @return null
+	 */
+	public function action_init_post_field() {
+	}
```
danielbachhuber: Do we need this method?

shortcode-ui.php
```diff
@@ -21,11 +21,14 @@
 
 require_once dirname( __FILE__ ) . '/inc/class-shortcode-ui.php';
 require_once dirname( __FILE__ ) . '/inc/fields/class-shortcode-ui-fields.php';
+require_once dirname( __FILE__ ) . '/inc/fields/class-field-attachment.php';
+require_once dirname( __FILE__ ) . '/dev.php';
```
danielbachhuber: Please remove this `/dev.php`

inc/fields/class-field-attachment.php
```diff
@@ -0,0 +1,77 @@
...
+			self::$instance->setup_actions();
+		}
+		return self::$instance;
+	}
+
+	private function setup_actions() {
+
+		add_filter( 'shortcode_ui_fields', array( $this, 'filter_shortcode_ui_fields' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ), 100 );
+		add_action( 'print_media_templates', array( $this, 'action_print_media_templates' ) );
```
danielbachhuber: Can we make sure these are only loaded when Shortcode UI is loaded?

bfintal:
+1 to this

roborourke:
+1, the fieldmanager one seems to be broken at the moment anyway

bfintal:
Tested, and works great :+1:

roborourke:
Nice one Matt. Works really well, cheers for adding in the library type feature too

mattheu:
Made a few updates - I think this is good to go.

danielbachhuber:
@mattheu We can probably ditch Fieldmanager media field then, no?

mattheu:
Yes - aside from backwards compatability with anyone using it already.

danielbachhuber:
Looks pretty solid. Only one minor issue #155

Issue #82, opened 2014-12-12
"Insert Element" should be disabled until a Post Element is selected


danielbachhuber:
See screenshot:

![image](https://cloud.githubusercontent.com/assets/36432/5416758/843740c8-81e7-11e4-846e-2db64abe042d.png)

danielbachhuber:
:shipit:

PR #147, opened 2015-02-12
Disable "Insert element" on shortcode selection screen

jitendraharpalani58:
See #82


danielbachhuber:
:money_with_wings: solid work!

PR #149, opened 2015-02-12
Fieldmanager templates not always output

roborourke:
In some cases a plugin might call `wp_enqueue_media()` and its action prior to `wp_enqueue_editor()` - this causes the `print_media_templates` action to run before the Fieldmanager addon can bind to it. The end result is the Fieldmanager templates and js don't get output.


danielbachhuber:
I'm :-1: on this change. See #153

danielbachhuber:
Reverted in 43447cc

Issue #100, opened 2015-01-02
Checkbox always checked


samhermes:
Once a checkbox is checked, it won't allow you to uncheck it. There doesn't seem to be any update to the shortcode if the box is unchecked, so the value remains true no matter what.

danielbachhuber:
Thanks for the report! I have a good sense of what's causing this — we'll try to get it fixed in the next release.

jitendraharpalani58:
@danielbachhuber @samhermes which checkbox is it? 

@samhermes can you please upload a screenshot?

danielbachhuber:
> which checkbox is it?

I think you'll need to create a sample checkbox field

danielbachhuber:
@samhermes Can you confirm the fix?

samhermes:
Confirmed. Thanks guys!

PR #150, opened 2015-02-13
Correct behavior for checkboxes

jitendraharpalani58:
See #100 

@danielbachhuber i have also update `refresh` method because it was throwing JS errors if you try to edit your shortcode. Because, `this.frame` no longer exist after you have inserted the shortcode.


jitendraharpalani58:
Related: https://github.com/fusioneng/fusion-theme/pull/1818

danielbachhuber:
> i have also update refresh method because it was throwing JS errors if you try to edit your shortcode.

Can you submit a PR fix for this next time? Thanks

PR #156, opened 2015-02-13
Fix JS error when attribute doesn't have `type`

danielbachhuber:
From #150


mattheu:
Would be good to have some smarter parsing of args - you can't not have a type - so we should throw an error or something.

danielbachhuber:
> you can't not have a type

`type` in this case is the `input` attribute, not a shortcode argument. Though, now that you mention it, we should be checking the shortcode UI registration argument cc @jitendraharpalani58

PR #157, opened 2015-02-13
Properly check module attribute instead of input attribute hack

danielbachhuber:
From #156


danielbachhuber:
cc @jitendraharpalani58

Issue #153, opened 2015-02-13
Decide whether Shortcake is coupled with TinyMCE or the media library


danielbachhuber:
The media library can be used independently of the editor, and vice versa.

We need more robust code for making sure Shortcake is only included for the editor

From #149

danielbachhuber:
I think Shortcake should be coupled with the editor, and make sure appropriate media assets are enqueued if the media library hasn't already been enqueued.

danielbachhuber:
@sanchothefat Can you look into a better solution for this, given that you were keen on fixing it?

mattheu:
At the moment we're pretty coupled to both. The UI is completely tied to the media library, and the rendering is completely coupled to the editor.

However - I think it actually makes more sense to couple with the media library - as you currenlty can't use without.

I can easily see the need to use the Shortcake UI outside of the editor. We pretty much have this requirement on Fusion at the moment. 

It would require more significant changes to use the UI outside of the Media Library -  however I see it as a definite possibility following https://github.com/fusioneng/Shortcake/issues/142

roborourke:
@danielbachhuber we're kind of at the mercy of WordPress core here, the `print_media_templates` action is coupled to the media library so we'd have to use `wp_enqueue_editor` and a generic hook like `admin_print_footer_scripts` instead.

danielbachhuber:
> I can easily see the need to use the Shortcake UI outside of the editor. We pretty much have this requirement on Fusion at the moment.

Huh? Where do previews appear, if not within TinyMCE?

danielbachhuber:
FWIW, #149 causes JS errors, so I need to revert.

danielbachhuber:
> we're kind of at the mercy of WordPress core here, the print_media_templates action is coupled to the media library so we'd have to use wp_enqueue_editor and a generic hook like admin_print_footer_scripts instead.

That's fine. I think we should ditch `print_media_templates` as it's been causing us troubles. See #139

roborourke:
Can you elaborate on the js errors you saw? I was getting the same problem hence the PR

danielbachhuber:
> Can you elaborate on the js errors you saw?

An internal JS file that has dependencies on Shortcake started throwing an undefined variable error.

I don't think the solution is to change the load action. I think the solution is to internalize the printing of our own templates, and also calling `wp_enqueue_media()` to make sure the media library assets already always available.

mattheu:
[...]

PR #158, opened 2015-02-13
Don't rely completely on wp_enqueue_editor or wp_enqueue_media

roborourke:
This is a first pass at moving to something less coupled to the editor or the media libraries. Once shortcake is fully decoupled from the media popup the `wp_enqueue_media()` can be removed from `Shortcode_UI\enqueue()`.

Fieldmanager is a bit annoying as the field classes available to shortcake have to be initialised whether shortcake is rendered or not but at least the templates aren't output. This is because Fieldmanager assumes all the field classes are created before the `admin_enqueue_scripts` action - `wp_enqueue_editor` is bound to `admin_print_footer_scripts` so there's no other way around it yet.

Addresses #153


shortcode-ui.php
```diff
@@ -58,3 +58,10 @@ function shortcode_ui_register_for_shortcode( $shortcode_tag, $args = array() )
 function shortcode_ui_get_register_shortcode( $shortcode_tag, $args = array() ) {
 	return Shortcode_UI::get_instance()->get_shortcode( $shortcode_tag );
 }
+
+/**
+ * Queue the shortcode UI scripts & templates manually
+ */
+function enqueue_shortcode_ui() {
```
danielbachhuber: Let's call this `shortcode_ui_enqueue_assets()` to properly namespace

roborourke: Sorted

roborourke:
@mattheu @danielbachhuber need me to do anything else on this or do you have another approach in mind?

danielbachhuber:
Well, I'd like to see them decoupled (e.g. media library code independent of TinyMCE views and vice versa). Not sure what @mattheu is currently thinking though.

roborourke:
@danielbachhuber at the moment this couples the output by default to the `wp_enqueue_editor` action as the plugin did originally but without relying on `print_media_templates`. It fixes a bunch of compatibility issues I've run into with things like the Black Studio MCE widget.

When someone needs to use the plugin in another context the `shortcode_ui_enqueue_assets()` function can be called so once the UI is decoupled from the media modal it should just continue to work using this function. We'd only need to remove the call to `wp_enqueue_media()` in the enqueue method.

mattheu:
Someone else has actually run into this exact same issue.

I think we should go ahead and merge this regardless as it fixes an issue. I'll take a look into decoupling the JS a bit further as a separate task.

inc/class-shortcode-ui.php
```diff
@@ -137,10 +155,12 @@ public function action_wp_enqueue_editor() {
 	 *
 	 * @return null
 	 */
-	public function action_print_media_templates() {
+	public function print_templates() {
```
danielbachhuber: Can we keep this name consistent?

danielbachhuber:
Sounds good. One minor nit.

inc/class-shortcode-ui.php
roborourke: Cheers @danielbachhuber, makes sense. Should be good now.

roborourke: D'oh - making this consistent with the action name it's hooked to rather than just reverting the name.

Issue #140, opened 2015-02-04
IE11 Crashes


Danny159:
There seem to be a problem with IE11.

When you edit a page that uses Shortcake and click to edit the shortcode IE11 freezes says a script it looping then crashes...

Is anyone else having this problem? Or know of a fix.

Thanks

danielbachhuber:
Sorry for the late reply. We can leave this open to see if anyone else can confirm.

mattheu:
Confirmed this bug - opening PR now

danielbachhuber:
Thanks for your work on this, @mattheu 

@Danny159 Can you confirm the fix?

PR #159, opened 2015-02-14
Refactor preview to prevent IE11 crashes

mattheu:
Shortcake crashes IE11 - https://github.com/fusioneng/Shortcake/issues/140

I narrowed it down to the preview - and I think the way it builds the iFrame. I wasn't that familiar with this bit of code so I dug into it a little bit. I ended up refactoring it a little bit.
- Use jQuery to build the iFrame instead of standard JS - since we're already using it.
- Removes need for template
- Now the code is much simpler - I think we can remove the iFrame constructor class.
- Use wp.ajax to fetch shortcode.

This fixes the IE11 crash


danielbachhuber:
@mattheu I get some errors:

image

Can you reproduce? Looks pretty good otherwise.

mattheu:
Merged with master and fixed a small error - however this was a different error from what you are seeing.

I tried in a few browsers, and on a couple of projects - and couldn't replicate your exact errors.

If you're still seeing this - would be good if you could you give me a bit more info.

danielbachhuber:
@mattheu were you able to fix the error?

mattheu:
I was unable to reproduce. I'll look into it some more today.

danielbachhuber:
I can't reproduce now. Let's ship to production and see if any users report.

Issue #139, opened 2015-02-02
Split templates into single templates and register via the WP Dependency API


franz-josef-kaiser:
As shown in [this PR](https://github.com/fusioneng/Shortcake/pull/118/files#diff-19950a22ee7d4e0bf323e00407199c1aR75) by @danielbachhuber the `.tmpl` scripts could be registered as _real_ scripts.

When split up into single files, this change allows the _easiest_ exchange of templates there could be. Just `wp_deregister_script( $handle );` the plugin template and `wp_register_script( ... );` your own extension.

The easiest and fastest way to load Underscore templates using the WP Dependencies API that I developed by now is [this sample plugin](https://gist.github.com/franz-josef-kaiser/6f5448d83558427c5e7d#). As you can see, it runs and fully supports everything core can deliver and is as close to core as it gets. I suggest altering the template registration to pretty much what is shown in the link.

If there is support for this change, I could file a PR in about a week or a bit more. Just ping me here.

franz-josef-kaiser:
We [now have a core ticket](https://core.trac.wordpress.org/ticket/31281) to move that one right in. In the meanwhile, the plugin could still incorporate it.

roborourke:
@franz-josef-kaiser main difficulty here is that for Fieldmanager at least the templates don't actually exist as files anywhere, it's generated code. Wondering if your example could be modified to allow registering a script as source code directly without need for an actual file

franz-josef-kaiser:
@sanchothefat Nice to read from you again (after .. years?) :)

Short question: Fieldmanager? wat?

roborourke:
@franz-josef-kaiser you too, has been a fe years now I guess!

So this is [Fieldmanager](https://github.com/alleyinteractive/wordpress-fieldmanager). Shortcake has built in support for a few of its field types if it detects the Fieldmanager plugin is installed and activated.

The code that outputs the templates is in this file https://github.com/fusioneng/Shortcake/blob/master/inc/fields/class-shortcode-ui-fields-fieldmanager.php#L48

It'd be cool if it was possible to enqueue the actual `<script>` tags without needing a physical file but perhaps that a bit crazy...

mattheu:
I don't think we should let the fieldmanager fields hold us back if we want to go down this route. We can always handle those separately.

franz-josef-kaiser:
> It'd be cool if it was possible to enqueue the actual `<script>` tags without needing a physical file but perhaps that a bit crazy

[...]

PR #161, opened 2015-02-16
Added color picker field

bfintal:
Added type `color`, uses WordPress' native color picker and created a new template for the new field.

I don't know where to put the initializing js though, so I added it to the template file. I also used `jQuery('[id="{{ data.attr }}"]')` instead of just specifying the ID via `#{{ data.attr }}` since that did not work when the shortcode is edited multiple times since the first initialization no longer exists on the succeeding call.

We need the color option for our: [Page Builder Sandwich](https://github.com/gambitph/Page-Builder-Sandwich)


inc/class-shortcode-ui.php
```diff
@@ -94,6 +94,9 @@ public function action_wp_enqueue_editor() {
 
 		do_action( 'shortcode_ui_loaded_editor' );
 
+		wp_enqueue_script( 'wp-color-picker' );
+		wp_enqueue_style( 'wp-color-picker' );
```
danielbachhuber: We should only load these if a colorpicker field type has been used.

inc/templates/edit-form.tpl.php
```diff
@@ -134,3 +134,20 @@
...
+		<# } #>
+	</div>
+	<script>
+		jQuery('[id="{{ data.attr }}"]').wpColorPicker({
+			change: function() {
+				jQuery(this).trigger('keyup');
+			}
+		});
+	</script>
+</script>
```
danielbachhuber: Same comment as above. It would be worth putting this in a separate file for this reason.

bfintal: How would you go about putting the template in another file? Probably need your intervention on how to go about splitting this one since as of the moment all the template files are in the same place

danielbachhuber: > How would you go about putting the template in another file?

There's discussion in #139

For the purposes of this pull request, at the very least we shouldn't call `wpColorPicker()` until the element is instantiated

bfintal: Would this mean that ideally I should follow what the Field Attachment #146 did and create my own `class-field-color.php` and put the template there as well?

danielbachhuber: Sure, I'm open to that as an interim solution

bfintal:
Revamped how the color field was included, now included the same way as #146 

Much cleaner now

js/field-color.js
```diff
@@ -0,0 +1,21 @@
+( function( $ ) {
+
+	var sui = window.Shortcode_UI;
+
+	sui.views.editAttributeFieldColor = sui.views.editAttributeField.extend( {
+
+		render: function() {
+			this.$el.html( this.template( this.model.toJSON() ) );
+			
+			jQuery(this.$el).find('input[type="text"]:not(.wp-color-picker)').wpColorPicker({
```
mattheu: Can just do `this.$el.find()` here

inc/fields/class-field-color.php
```diff
@@ -0,0 +1,94 @@
...
+		wp_enqueue_script( 'shortcake-field-color', $script, array( 'shortcode-ui' ) );
+		
+	}
+
+	/**
+	 * Enqueue the color picker only when there's a color option used
+	 */
+	public function load_color_picker() {
+		
+		foreach( Shortcode_UI::get_shortcodes() as $shortcode ) {
```
mattheu: I really like this conditional loading of the JS - but this isn't working for me.

I think you need to use `Shortcode_UI::get_instance()->get_shortcodes()` here.

Also - why not move loading of the templates and the `shortcake-field-color` script here as well?

mattheu:
Looking great

One small comment on the styling. This rule `.media-frame a { border-bottom: none }` in `media-views.css` is removing the border - which is a little annoying.

I think the only solution is to add a style to the shortcake CSS to fix this

bfintal:
@mattheu all done :) Also added in the attribute checker. Might be beneficial if that function was ported over to the `Shortcode_UI` class so other fields can use it.

bfintal:
@mattheu Good catch on the bottom border, I must be going blind :+1:

danielbachhuber:
:+1: Thanks for your work on this, @bfintal

mattheu:
> Might be beneficial if that function was ported over to the Shortcode_UI class so other fields can use it.

+1 see #173

Issue #142, opened 2015-02-06
Abstract shortcode-ui.js into modules with Browserify


danielbachhuber:
`shortcode-ui.js` is getting quite long and complex. We should break it into logical modules using Browserify.

@noppanit Can you weigh in with some thoughts in this regard?

noppanit:
I'd like to refactor this and add some tests as well. Or @jitendraharpalani58 if you want to do it, it's ok for me as well.

danielbachhuber:
@jitendraharpalani58 wants to work on it, but you'd make a great review partner :)

danielbachhuber:
:ship:

PR #162, opened 2015-02-16
Abstract Shortcode JS to MVC

jitendraharpalani58:
See #142 

@noppanit Can you help me setup `browserify` when you get a chance ?


noppanit:
@jitendraharpalani58 I believe it's fixed already. Please try again. Also you need to run `npm install` again to install all the dependencies. I had to fix version of Remapify because the issue in the latest version.

jitendraharpalani58:
@danielbachhuber 
Abstraction is done. 

One thing which is remaining is if someone wants to extend `shortcake` functionality, will they be using browserify? For example, like we did for `select2ui` by extending `sui.views.editAttributeField`. 

There are two ways
- I can still store everything in `sui = window.Shortcode_UI = {}` like we had earlier
- User may choose to use browserify and require the view they want to extend. 
  in our case it would be something like: 

```
var EditAttributeField = require('sui-views/edit-attribute-field');

var blah = EditAttributeField.extend...
```

What do you  suggest?

jitendraharpalani58:
@mattheu @noppanit 
Can you guys review this PR please.

danielbachhuber:
@jitendraharpalani58 Can you merge master so this merges cleanly?

jitendraharpalani58:
@danielbachhuber @mattheu @noppanit 
I added `sui` global variable. I am still using same naming conventions for the constructor as they were to support backward compatibility. Otherwise, it will break for the users who are using our plugin. 

Please review. This should be good to push live.

Issue #93, opened 2014-12-22
Support for editing shortcode inner content


danielbachhuber:
If I have a shortcode like this:

```
[my-shortcode]Some awesome inner content[/my-shortcode]
```

It should be possible to edit that inner content.

mattheu:
This is currently possible. Just add an attribute as normal with the name 'content'.

Perhaps this should be a bit more explicit....

danielbachhuber:
> Just add an attribute as normal with the name 'content'.

Seems a bit magical — what if I had an attribute named "content"?

mattheu:
Yeah agreed.

jitendraharpalani58:
@danielbachhuber This issue is a bit unclear to me. We need add a new feature to edit content. 
Consider this for example: 

```
'lede' => array(
            'label' => 'Lede',
            'listItemImage' => 'dashicons-editor-textcolor',
            'attrs' => array(
                array(
                    'label' => 'Text',
                    'attr'  => 'content',
                    'type'  => 'textarea',
                ),
            ),
        ),
```

can we add another element with key as content

```
'lede' => array(
            'label' => 'Lede',
            'listItemImage' => 'dashicons-editor-textcolor',
                       'content' => array( 
                               'type' => 'String', // this can be reference/original another shortcode to support nesting. 
                               'value' => 'My test content',
                        ),
        ),
```

Am I thinking it right?

mattheu:
Yes I think that would make things a bit clearer. Currently we just treat an attribute with the attr 'content'  as a speacial case.

danielbachhuber:
We should remove `content` as a special case, and handle inner content in a proper manner.

jitendraharpalani58:
@danielbachhuber what are you proposing ? 
I think having a `content=>{{value}}` pair is a good way to go as it gives an option to support nested shortcodes. Along with this, we can add some front end UI also to support inner content (to make it more intuitive). 

So there will be two ways of adding inner content
- via code (`content=>{{value}}`). Default value: true (will show the add/edit button)
- Add/edit content button on UI.

danielbachhuber:
The current problem is that `content` is a valid shortcode attribute, but we can't use it. We should have a special flag for it instead.

PR #164, opened 2015-02-17
Allow to edit content - first pass

jitendraharpalani58:
See #93 

@danielbachhuber This is what I was thinking.


inc/class-shortcode-ui.php
```diff
@@ -35,9 +35,10 @@ private function setup_actions() {
 	public function register_shortcode_ui( $shortcode_tag, $args = array() ) {
 
 		$defaults = array(
-			'label'         => '',
-			'attrs'         => array(),
-			'listItemImage' => '',   // src or 'dashicons-' - used in insert list.
+			'label'             => '',
+			'attrs'             => array(),
+			'listItemImage'     => '',   // src or 'dashicons-' - used in insert list.
+			'shortcode_content' => true,
```
danielbachhuber: Let's call this `inner_content`. I think `shortcode_` is implied.

inc/templates/edit-form.tpl.php
```diff
@@ -134,3 +134,16 @@
...
+
+<script type="text/html" id="tmpl-shortcode-ui-content">
+	<div class="field-block">
+		<div class="content-display-container">
+			<p class="content-display">{{ data.shortcode_content }}</p>
+			<a href="#" class="button-edit">Edit Content</a>
+		</div>
+		<div class="content-edit-container" style="display: none;">
+			<textarea class="content-edit">{{ data.shortcode_content }}</textarea>
+			<a href="#" class="button-save">Save Content</a>
```
danielbachhuber: Why do you have "Edit" and "Save"?

jitendraharpalani58: I think it makes sense to edit content, if the need be, directly from shortcake UI.

inc/class-shortcode-ui.php
jitendraharpalani58: will do

inc/templates/edit-form.tpl.php
danielbachhuber: Sure — why not just dump it in a textarea then?

jitendraharpalani58: Looks more presentable.

mattheu: We don't use the edit/save UX for other fields. I'm not sure that inner content should be different. I think it would be nicer if it just live updated like the other fields.

We can probably re-use the same views we use for other fields. It might be nice if it was customizable too - as a textarea might not always be appropriate.

jitendraharpalani58: @mattheu Can you think of any use cases where `inner_content` will not be plain text (textarea), neither a combination of nested shortcodes ?

mattheu: Not really - but I could perhaps see that either a simpler text input or a more complex TinyMCE editor might be useful.

jitendraharpalani58:
@danielbachhuber Preview is not working anymore :( I am not able to understand what is the purpose of 
`sui.utils.shortcodeViewConstructor`. Lets talk about this today.

mattheu:
Preview got broke, but should be fixed on master - might need to merge this.

mattheu:
I can see that `sui.utils.shortcodeViewConstructor` is confusing - and can probably be improved.

WordPress can already handle rendering a shortcode preview using the MCE view API. You can add a preview for any shortcode using- `wp.mce.views.register` - it requires 2 args - the shortcode tag and the 'constructor'. The constructor should have 
- View property - which is used when creating a `wp.mce.View`.  
- Edit method - to handle what happens when the preview is clicked.

This is not well documented but can be found in `wp-includes/js/mce-view.js`.

We are registering a MCE view for all shortcodes added to shortcake. 

This preview works by making an ajax request to get the rendered shortcode. So the view object in `sui.utils.shortcodeViewConstructor` handles this.

The edit method has to convert the shortcode string back to a shortcake shortcode model, and open the edit form.

sui.utils.shortcodeViewConstructor is the constructor used by each shortcake shortcode when registering the MCE view.

jitendraharpalani58:
@mattheu

> Preview got broke, but should be fixed on master - might need to merge this.

It worked! Thanks!! 

Going through your notes about `sui.utils.shortcodeViewConstructor`

jitendraharpalani58:
@mattheu

can you please take a look why `inner_content` is not being saved. I am not able to figure out whats going on.

js/shortcode-ui.js
```diff
@@ -837,11 +868,11 @@ var Shortcode_UI;
...
 
-					if ( attr.get( 'attr' ) === 'content' && ( 'content' in options.shortcode ) ) {
-						attr.set( 'value', options.shortcode.content );
-					}
-
 				});
+				
+				if ( 'content' in options.shortcode ) {
+					shortcode.set( 'content', options.shortcode.content )
+				}
```
mattheu: I think this should be:

``` php
if ( 'inner_content' in options.shortcode ) {
    shortcode.set( 'inner_content', options.shortcode.content )
}
```

mattheu:
Also `content` needs updating to `inner_content` here: 

``` php
if ( matches[3] ) {
    currentShortcode.set( 'content', matches[3] );
}
```

see `sui.utils.shortcodeViewConstructor.edit`

mattheu:
Another thing that might be neater - is if you could actually reuse the editAttributeView for the content. I was thinking something like this:

``` php
var contentAttributeModel = new sui.models.ShortcodeAttribute( {
    attr:  'inner_content',
    labell: 'Content',
    type:  'textarea',
    value: t.model.get('inner_content'),
} );

// Add field for inner_content
var view = new sui.views['editAttributeField']( { model: contentAttributeModel } );
view.template = wp.media.template( 'shortcode-ui-field-textarea' );
view.shortcode = t.model;

// Update changes.
contentAttributeModel.on( 'change:value', function() {
    t.model.set( 'inner_content', contentAttributeModel.get( 'value' ) );
} );

t.views.add( '.edit-shortcode-form-fields', view );
```

This would allow you to remove the content model and template.

mattheu:
One other consideration. Not all shortcodes support content. It should be possible to enable/disable this field.

danielbachhuber:
> It should be possible to enable/disable this field.

Right. And it should be disabled by default.

On our call today, @jitendraharpalani58 and I discussed supporting nested shortcodes. In the interest of simplicity, let's treat nested shortcodes the same as any inner content — a string. We can maybe revisit later.

jitendraharpalani58:
@danielbachhuber If we are treating `inner_content` as String for now, I have kept it real simple. 
- Default to false
- If `inner_content` attribute is present, but empty, show empty `textarea`
- If `inner_content` attribute is present, and not empty, populate `textarea` with that value

If/when we decide to add more features, we can create a `inner_content` model.

danielbachhuber:
I think `'inner_content'     => 'false'` is a hack and that we should do something better by default.

At the PHP-level, I'd expect to be able to register `inner_content=>true` or, if I wanted to change the label or have some default value, something like `inner_content=[label=>My Content,default_value=>Apples]`

If `inner_content=true`, the PHP API would fill in the values like this:

```
// inner_content=true is a valid argument, but we want more detail
if ( is_bool( $args['inner_content'] ) && true === $args['inner_content'] ) {
    $args['inner_content'] = array(
        'label'            => esc_html__( 'Inner Content', 'shortcode-ui' ),
        'description'      => '',
        'placeholder'      => '',
        );
}
```

There's no reason to have an extra template either — we can reuse the `textarea` template.

Another important behavior should be to persist existing `inner_content` for a shortcode if the shortcode has it but no UI was registered. It would be nice to have tests for this behavior, which probably means #162 should land before this.

jitendraharpalani58:
@danielbachhuber will inner_content be backward compatible for the users who have already used `attr=>'content'`?

jitendraharpalani58:
@danielbachhuber final commit resolves all the issues. 
@mattheu thanks for your help with this one. 

Please review.

inc/class-shortcode-ui.php
```diff
@@ -35,13 +35,44 @@ private function setup_actions() {
...
+		// inner_content=true is a valid argument, but we want more detail
+		if ( is_bool( $args['inner_content'] ) && true === $args['inner_content'] ) {
+			$args['inner_content'] = array(
+				'label'            => esc_html__( 'Inner Content', 'shortcode-ui' ),
+				'description'      => '',
+				'placeholder'      => '',
+			);
+		}
+		
+		$num_attrs = count( $args['attrs'] );
```
danielbachhuber, line 58: Can we have some mention that this is a backwards compatibility shim that we'll be removing in the next version?

jitendraharpalani58, line 58: adding that now.

inc/templates/edit-form.tpl.php
```diff
@@ -52,7 +52,9 @@
 <script type="text/html" id="tmpl-shortcode-ui-field-textarea">
 	<div class="field-block">
 		<label for="{{ data.attr }}">{{ data.label }}</label>
-		<textarea name="{{ data.attr }}" id="{{ data.attr }}" placeholder="{{ data.placeholder }}">{{ data.value }}</textarea>
+		<textarea name="<#if ( data.attr ) { #>{{ data.attr }}<#}else{#>inner_content<#}#>"
```
danielbachhuber: Hm, does this mean I couldn't have `inner_content` as an attribute? Maybe we do need a separate template then.

js/shortcode-ui.js
```diff
@@ -118,7 +118,19 @@ var Shortcode_UI;
...
 		},
 	});
-
+	
+	/**
+	 * Shortcode Attribute Model.
+	 */
+	sui.models.InnerContent = Backbone.Model.extend({
+		defaults: {
+			label:       'Inner Content',
```
danielbachhuber: This default value should be our translatable string.

inc/templates/edit-form.tpl.php
jitendraharpalani58: the whole purpose of adding `inner_content` is to not use it as attribute. That was the reason i added a new template for `inner_content`. Should I revert back to adding a new template ?

js/shortcode-ui.js
jitendraharpalani58: then, I can just leave it empty in JS?

inc/templates/edit-form.tpl.php
danielbachhuber: > Should I revert back to adding a new template ?

Yes please. Apologies for the misdirection

jitendraharpalani58:
@danielbachhuber made changes according to your comments.

danielbachhuber:
@jitendraharpalani58 Can you move this to the new Browserify pattern? Feel free to create a new branch and PR if you need to.

jitendraharpalani58:
Will do

jitendraharpalani58:
@danielbachhuber this is ready to push live!

PR #168, opened 2015-02-17
167 - preview completely broken

### LINKED ISSUE #167: Preview is broken entirely right now
image

@mattheu thinks #143

mattheu:
`do_shortcode` ajax request was changed to use `wp_send_json_success` to return the data. This means that the response is now an object with the HTML string passed as the data property.

The preview was not updated accordingly.

Fixes #167


danielbachhuber:
:+1:

Issue #116, opened 2015-01-28
Ability to edit a shortcode's attributes inline


danielbachhuber:
For shortcodes that don't have substantially complex fields, it might be nice to be able to edit a shortcode's attributes inline.

sc0ttkclark:
Such yes

bfintal:
This would be great, although I'm not so certain that TinyMCE views can readily handle this

mattheu:
I think a first step could be to allow shortcodes to override the editor form view and provide their own. 

This way - you could completely customize the appearance. You could make the form look exactly like the rendered shortcode.

The next step - you could provide an option to render the form inline instead of the preview for a given shortcode. Providing editable forms doesn't totally break the MCE views this should be OK.

danielbachhuber:
Inline editing [came up](https://make.wordpress.org/core/2015/06/29/shortcake-shortcode-ui-chat-summary-june-29th-2015/) as a core integration point. We should prioritize it for v0.6.0, and explore how other editors do this / content blocks.

Another consideration for inline editing is block vs. non-block shortcodes. Inline editing could be less functional for non-block shortcodes, as we'll have less space to work with.

cliffordp:
:+1:
Right now the only way is to go to the "Text" editor instead of the "Visual" editor... but then you lose your place (which [xyz] shortcode was I wanting to edit? where in this long post was that [abc] shortcode? etc)

And possibly related: https://github.com/fusioneng/Shortcake/issues/115 (select existing text and turn into a shortcode)

danielbachhuber:
Some [conversation from Slack](https://wordpress.slack.com/archives/feature-shortcode/p1444072823000031):

> goldenapples [12:22 PM] Only way I can see that happening, is allowing shortcodes to render an edit preview with blocks of text marked as “contenteditable”, and using some postmessage magic to maintain the link between those fields in the preview and the shortcode itself.
> goldenapples [12:23 PM] It seems possible, but not easy or obvious, and even after exploring technical feasibility probably needs a lot of UX thought.
> danielbachhuber [12:23 PM] there’s another approach we could take: open form fields inline instead of modal
> danielbachhuber [12:24 PM] I think you’d get 80% of the benefit for 20% of the effort
> goldenapples [12:24 PM] hmm. what does “inline” mean in that context? like, rendering form inputs inside the wpview iframe?
> danielbachhuber [12:25 PM] yeah
> danielbachhuber [12:25 PM] or on top of
[...]

PR #174, opened 2015-02-19
Template Filter for overriding views

bfintal:


danielbachhuber:
@bfintal Could you describe in more detail what you're trying to do?

mattheu:
This would allow you to provide completely new templates if you want. I agree this would be nice - but we've actually been discussing an alternative different approach here: https://github.com/fusioneng/Shortcake/issues/139

I have a slightly different use case however. It would be good to specify a completely custom `EditShortcodeForm` view on a per shortcode basis. Being able to have a completely custom interface for a single shortocode would be a nice first step towards inline editing.

danielbachhuber:
> It would be good to specify a completely custom EditShortcodeForm view on a per shortcode basis. Being able to have a completely custom interface for a single shortocode would be a nice first step towards inline editing.

This seems like a reasonably substantial architectural change. We should discuss as a part of #116

PR #178, opened 2015-02-20
Do template after default args are set so default args are applied

mattheu:
The template is being rendered _before_ the default args are set. So if you don't pass a custom string for the add attachment button, the button is empty.

Simple fix - render the template after the default args are set.


danielbachhuber:
We need some tests :)

Issue #171, opened 2015-02-18
Add support for wpautop()


nturbanov:
Since there is no possibility to add a TinyMCE field yet, support for wpautop() would be great, ie. a single line break renders a `<br />` and a double line break a `<p>`.

`<?php echo wpautop( $content ); ?>` works in the front-end and the WP editor, but the html tags are visible in the Shortcake textarea, except the very fist opening `<p>` and the very las closing `</p>`.

danielbachhuber:
There's a couple ways you could hack this right now:
1. Hook into the preview actions to add a `<p>` tag before and after.
2. Have some conditional logic in the shortcode that runs the result through `wpautop()` when the context is admin ajax.

I suppose the easiest way to do this would be to run the preview HTML through a filter. It seems messy to have two actions _and_ a filter though.

@mattheu Thoughts?

mattheu:
This is a tricky one.

So I think that this should be handled in the JS as thats where the problem lies - rather than adding a filter.

I _think_ we can fix this by running the content through `switchEditors._wp_Nop` in 2 places. Once when outputting the formatted shortcode. Once when parsing the content from the raw shortcode.

I'm going to open a PR with this - but I am a little concerned that it have other unforseen consequences...

danielbachhuber:
Moving this out of v0.3.0

PR #179, opened 2015-02-20
Run content through _wp_Nop

mattheu:
See #171

A few concerns here
- Should this always run or should we set a flag?
- Should we really be using _wp_Nop?
- Is this a situation where it would be good to have unit tests? Maybe we should start some.


danielbachhuber:
> Should this always run or should we set a flag?

I think we need to always run, or not run. Decisions not options ;)

> Should we really be using _wp_Nop?

I'm not convinced. Seems like we'd make the code more brittle by being dependent on `switchEditors`

> Is this a situation where it would be good to have unit tests?

Yep

mattheu:
After doing some further testing - I think its fine to just always run this.

Agreed that we don't want to depend on switchEditors, _wp_Nop is a pretty hacky ~100 lines of regex that I don't really want to duplicate - and I guess that we would be able to support fewer edge cases.

I've written some simpler regex to do this - should be sufficient. I'll look into doing some tests for this before we merge though.

danielbachhuber:
> Agreed that we don't want to depend on switchEditors

Can we make it fail gracefully then if the variable isn't there?

mattheu:
Sure. I think if we go down the route of supporting the WP editor inside shortcake ui, and also nested shortcodes, we want this to be pretty robust and I'd rather lean on the built-in WP Core functionality where possible

mattheu:
I've synced this up with master and added some of the things we discussed.

However this is no longer working with WP trunk. The problem is that MCE views no longer  work at all if the shortcode content has HTML. I think this broke here: https://github.com/WordPress/WordPress/commit/b799bd8e3385373a823df72f472a23a501f694e2

I'll look into how feasable this is to support.

mattheu:
Support for HTML and line breaks just got fixed in core. See https://core.trac.wordpress.org/changeset/32678 

A somewhat simpler solution to https://github.com/fusioneng/Shortcake/pull/322 - although it only works for the inner content, and not data stored in attributes.

I think this should go in regardless of #322, and we can review whether we need to extend support for HTML to all attributes.

PR #180, opened 2015-02-21
Numeric attribute fix

### LINKED ISSUE #128: Support positional (unnamed) parameters
Some shortcodes accept positional parameters. e.g. [bw_link github.com/fusioneng/Shortcake/issues/128]
which would display a link to this issue.

Some shortcodes accept multiple positional parameters
e.g. [wp v p] - which would display WordPress 4.1 PHP 5.5.18

Positional parameters appear in the $atts array as unkeyed values.
So, for the above,
- when $atts[0] == 'v' then the shortcode displays the WordPress version
- when $atts[1] == 'p' then the shorcode display the PHP version

Some shortcodes even accept these unnamed parameters in any order.

The UI should be able to detect positional parameters and create the shortcode without the parameter name.

[bw_dash 0=admin-links 1=x5 something] doesn't work as expected. $atts[0] is not set. 

Note: WordPress shortcode processing currently appears to lose the value for a parameter named 0, but does pass other numeric parameters; more by luck than judgement.

bfintal:
Fixes #128

Doing this

```
'attrs' => array(
    array(
       'label' => 'url',
       'attr'  => '0',
       'type'  => 'textarea',
    ),
),
```

creates a shortcode with an attribute `0="value"`, WordPress cannot parse this as an unnamed attribute. The regex here removes numeric attributes and removes them. This doesn't affect attribute-like strings outside the shortcode definition and doesn't affect normal string attributes.


danielbachhuber:
> The regex here removes numeric attributes and removes them.

The problem is that it's also special-casing numeric attributes, so I can no longer use `0` as an attribute (as unlikely as that is).

bfintal:
You can, unnamed attributes in WP are placed as numeric keys in the `$attr` variable during shortcode parsing. So I from what I can tell unnamed & numeric are treated the same way in WP. So there shouldn't be any problems.

Tested this using Jetpack's **googlemaps** shortcode:

```
/**
 * [googlemaps] shortcode
 *
 * Example usage:
 *   [googlemaps http://maps.google.com/maps?f=q&hl=en&geocode=&q=San+Francisco,+CA&sll=43.469466,-83.998504&sspn=0.01115,0.025942&g=San+Francisco,+CA&ie=UTF8&z=12&iwloc=addr&ll=37.808156,-122.402458&output=embed&s=AARTsJp56EajYksz3JXgNCwT3LJnGsqqAQ&w=425&h=350]
 *   [googlemaps https://mapsengine.google.com/map/embed?mid=zbBhkou4wwtE.kUmp8K6QJ7SA&w=640&h=480]
 */
function jetpack_googlemaps_shortcode( $atts ) {
    if ( !isset($atts[0]) )
        return '';

    $params = ltrim( $atts[0], '=' );

    var_dump( $atts[0] );

    // didn't paste the rest of the code
```

The `var_dump( $atts[0] )` there returned the correct attribute after the `preg_replace` ran.

mattheu:
Also related - https://github.com/fusioneng/Shortcake/issues/128

I'm not sure how we would tackle this.

EDIT - Oh - I see  you already commented on this.

bfintal:
Did some tests and it seems that this PR just solves part of the problem. Will get back with an explanation.

danielbachhuber:
I think we should have some unit test coverage here.

bfintal:
After extensive testing and debugging, I'm revamping my fix for this. I think this is the right way to approach this.

The **real** problem: _unnamed attributes are not recognized by the editor views & the UI_

I removed my initial commit with the regex, and now the views recognize the unnamed attributes.

The fix is to handle unnamed attributes as numeric ones. That means:

```
array(
   'label' => '',
   'attr'  => '0',
   'type'  => 'text',
   'value' => 'value1'
),
array(
   'label' => '',
   'attr'  => '1',
   'type'  => 'text',
   'value' => 'value2'
),
```

Will now generate:

`[shortcode "value1" "value2"]`

WordPress/PHP-wise, these attributes are accessible using the normal `$attr[0]` and `$attr[1]`

The fix is not yet complete. The only thing missing now is that editing an existing shortcode turns the attribute input fields blank, like the values are not being found.

bfintal:
The last part of the solution may be is to add another regex at the function `sui.utils.shortcodeViewConstructor.edit` to check for unnamed attribute values, then do an `attr.set` on those. TBH, I'm having trouble with figuring out a clean regex for this. Will continue trying in the morning if nobody solves this :)

bfintal:
Alright, finally finished the fix. I'm very well pleased with the results.

bfintal:
Closing and doing a new cleaner PR #198, since this now has tons of conflicts.

mattheu:
>  now has tons of conflicts.

We just landed some big changes to the structure of the JS. Apologies for the conflicts!

bfintal:
> Apologies for the conflicts!

No problem! It's about time I tried to learn how to use grunt.

PR #182, opened 2015-02-23
Moved label for checkbox

bfintal:
Currently there's no label beside the checkbox. I propose moving the existing label beside the checkbox.


mattheu:
Works great. Thanks!

PR #186, opened 2015-02-24
Fix fatal error when post is not present

### LINKED ISSUE #185: Failed to load preview when editing in non-post type
For example, this plugin uses TinyMCE inside Appearance > Widgets: https://wordpress.org/plugins/black-studio-tinymce-widget/

Views always show a "Failed to load preview" because post data cannot be set up. This should work even without post data.

This was reported to us https://github.com/gambitph/Page-Builder-Sandwich/issues/50

bfintal:
When the visual editor is used outside a post/page, Insert Post Element stops outputting previews. Post data shouldn't really be required to output shortcode previews in the editor.

Fixes: #185

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -164,9 +164,11 @@ class Shortcode_UI {
 			exit;
 		}
 
-		global $post;
-		$post = get_post( $post_id );
-		setup_postdata( $post );
+		if ( ! empty( $post_id ) ) {
+			global $post;
+			$post = get_post( $post_id );
+			setup_postdata( $post );
+		}
 
 		ob_start();
 		do_action( 'shortcode_ui_before_do_shortcode', $shortcode );


mattheu:
Could you provide some more info on how to recreate the fatal error? I tried to recreate by creating a `wp_editor` on an options page. But it all worked fine.

One caveat is that if your shortcode only works within the context of a post. Currently we hack this by setting up the post object/postdata using the ID found in the `#post_ID` field. 

If you're using it outside of the post context, BUT your shortcode needs this data, the only way round is to manually add this hidden field.

bfintal:
A user used this plugin: https://wordpress.org/plugins/black-studio-tinymce-widget/ with our Page Builder Sandwich plugin.

The first plugin creates a new widget called Black Studio. That contains a TinyMCE editor as one of the settings in the widget. 

Since the "Add Media" button is there while editing the Black Studio widget in Appearance > Widgets, the shortcode UIs are available also. Insert that widget then try placing a shortcode. It will fail to load the preview. The error can be seen in the developer console when you check out the responses of the ajax calls made to generate the shortcode view.

mattheu:
I still can't reproduce. I get a load of JS errors when I try to use this plugin. Your PR doesn't fix it.

The errors are actually in WP core JS files - and deleting all our JS doesn't fix. So the issue must be due to the order in which scripts are loaded.

```
media-audiovideo.js:90 Uncaught TypeError: Cannot read property 'id' of undefined
media-views.js:2656 Uncaught TypeError: Cannot read property 'audio' of undefined
```

Digging through the  Black Studio TinyMCE editor code I can see that they're calling  `do_action( 'wp_enqueue_editor', array( 'tinymce' => true ) );` on the `admin_print_scripts` action. This is earlier than WP does it - core does it in the footer.

Actually - I don't think this line is needed at all. They are correctly using `wp_editor` which should automatically handle all the required scripts. We're making use of this to ensure our scripts are only loaded when the editor is present.

If you try removing this line - it all works fine. 

I think this PR is reasonable - only set up post data when an ID is passed. I'm happy to merge. But I don't think it fixes your problem.

bfintal:
Here's how I did it http://cl.ly/3O2F133X1T13/ScreenFlow.mp4

I'm using the Black Studio plugin along with the latest master Shortcake code and the pullquote sample

mattheu:
Hmm - the video doesn't seem to work for me

mattheu:
Oh... hang on - it does - only chrome didn't like it.

I'm going to go ahead and merge this. Even though I was unable to reproduce - it makes sense to only set up the post data if it is actually passed.

danielbachhuber:
@mattheu When you merge PRs, please make sure they're assigned to a milestone with one of the labels. It makes release notes 100x easier to produce. Thanks

jakejackson1:
I have a similar issue with `wp_editor` and Gravity Forms. The error coming back from AJAX is "Something&#039;s rotten in the state of Denmark".

I debugged it to the class-shortcode-ui.php file's `if ( ! current_user_can( 'edit_post', $post_id ) || ! wp_verify_nonce( $_POST['nonce'], 'shortcode-ui-preview' ) ) {`. 

The `! current_user_can( 'edit_post', $post_id )` section appeared to be the issue here (in this case the post ID is null so the check seems to always fail).

PR #189, opened 2015-02-24
Allow html in descriptions

bfintal:
This is so that you can display links, strong, em, images, etc html
tags in descriptions. Currently, these are being escaped.


danielbachhuber:
I like the idea, but I'm not sure how VIP is going to feel about this.

bfintal:
Oh sorry, I don't know about that one :(

bfintal:
What isn't allowed here? Perhaps the description parameter can be run against `wp_kses_allowed_html( 'post' );` first, that would be safer and should prevent any malicious stuff.

danielbachhuber:
> What isn't allowed here?

Primarily `<script>`

> Perhaps the description parameter can be run against wp_kses_allowed_html( 'post' );

Escaping needs to happen as close to output as possible. The problem is with removing auto-escaping from the underscores template

bfintal:
@danielbachhuber Now strips unwanted html from descriptions. The only allowed html tags are the ones specified by `wp_kses_allowed_html( 'post' )`

mattheu:
Yeah this is a tough one. It would be nice to support at least a very basic set of HTML here - but I would definitly prefer to continue using the escaping built in to underscores templates.

bfintal:
The PR allows these: https://gist.github.com/bfintal/ba1277bd7bc04a7c64b9

danielbachhuber:
@bfintal Can we pass the allowed tags to Underscores, and still have Underscores perform the escaping?

bfintal:
@danielbachhuber I don't think it would allow it. Underscores would escape the `<` and `>` and those would appear as `&lt;` and `&gt;` in the description.

danielbachhuber:
This was addressed in #499

PR #196, opened 2015-02-24
Fix misnamed variable `Backbonen`

danielbachhuber:
Having it misnamed doesn't break preview though. Do we need it localized?


mattheu:
I guess `Backbone` is available globally as its already loaded in the admin.

Issue #195, opened 2015-02-24
Shortcode previews dependent on JS are broken


danielbachhuber:
Caused by #143 

Shortcodes dependent on JS need to be sandboxed in an iframe so document.ready fire.

danielbachhuber:
@mattheu You dropped https://github.com/fusioneng/Shortcake/pull/143/files#diff-0d42b6c6c12ff49c12fe9b436c6e95b0L861 entirely but it's not clear why — I don't think core's code gives us an equivalent.

PR #197, opened 2015-02-24
Fix shortcode previews with JS by restoring code that sandboxes into an iframe

danielbachhuber:
See #195


danielbachhuber:
Errors :(

danielbachhuber:
image

danielbachhuber:
Caused by `iframeDoc = iframe.contentDocument.document;`

PR #198, opened 2015-02-25
Unnamed/numeric attribute fix (revised)

bfintal:
This is a continuation of the fix for unnamed/numeric attributes. This PR replaces #180


js/build/shortcode-ui.js
```diff
@@ -167,7 +167,15 @@ var Shortcode = Backbone.Model.extend({
 			if ( attr.get( 'attr' ) === 'content' ) {
 				content = attr.get( 'value' );
 			} else {
-				attrs.push( attr.get( 'attr' ) + '="' + attr.get( 'value' ) + '"' );
+				
+				// Numeric attribute names
+				if ( ! isNaN( attr.get( 'attr' ) ) ) {
+					attrs.push( '"' + attr.get( 'value' ) + '"' );
```
danielbachhuber: Do we need to wrap in quotes?

danielbachhuber:
@mattheu How do you feel about this?

bfintal:
@danielbachhuber removed the quotes

js/build/shortcode-ui.js
```diff
@@ -167,7 +167,15 @@ var Shortcode = Backbone.Model.extend({
 			if ( attr.get( 'attr' ) === 'content' ) {
 				content = attr.get( 'value' );
 			} else {
-				attrs.push( attr.get( 'attr' ) + '="' + attr.get( 'value' ) + '"' );
+				
+				// Numeric attribute names
+				if ( ! isNaN( attr.get( 'attr' ) ) ) {
```
mattheu, line 195: Lots of double negatives here... We could check `$.isNumeric( attr.get( 'attr' ) )` instead?

mattheu:
Whilst I'd prefer to just not support numeric/positional attributes for simplicity - it is a feature of shortcodes. :smile: 

This works well - I think this its the best solution to the problem.

Its a bit out of sync with master - but shouldn't be too bad to update.

mattheu:
A couple of issues: 

I suppose the key is in the word 'numeric'... but should we also account for users who specify the shortcode attribute name as an Int?

If not all numeric attributes are supplied it should probably default to `false|null|` - to avoid getting out of sync. (eg - shortcode has 2 numeric attrs. Edit using shortcake and remove value of attr 1. Update. Value of 2 is now passed first...

bfintal:
@mattheu I believe handling of numeric attributes is the same. `[myshortcode 1="foo"]` is used as `$attr[1]` in the shortcode function. Take note that in WP, attribute `0` doesn't work.

I did the change. Empty unnamed attributes now appear as `false` when left blank in order to preserve succeeding unnamed attributes.

bfintal:
Synced the code to the current master

danielbachhuber:
@mattheu how do you feel about this?

danielbachhuber:
@bfintal One more sync?

Bump @mattheu

bfintal:
Hold on, things aren't working as they are supposed to

danielbachhuber:
This will need to be handled in a later release.

PR #200, opened 2015-02-25
Fix iframes 2

mattheu:
The issue is that shortcodes that contain JS were not getting sandboxed. #195

This is a half-revert of #143 - and fixes the immediate issue.


danielbachhuber:
Thanks @mattheu — really appreciate it.

Out of curiosity, did you figure out why my PR didn't work?

danielbachhuber:
This PR has a bug where the shortcode is added again if you edit it

danielbachhuber:
But this is just for content that's been iframed.

mattheu:
These damn iframes...

I tried reverting back to 18f2bdde3a13e64fbd5df31b48c67c917b1caa90 to check it was working then - and I think this problem has been present all along.

danielbachhuber:
Hm, pretty sure we had fixed it?

danielbachhuber:
> and I think this problem has been present all along.

Just tested again. Guess it was a bug before :( #202

Thanks for your work on this, @mattheu

Issue #181, opened 2015-02-22
TinyMCE views API improvements for 4.2


ellatrix:
Hi,

Just wanted to say that I'm redoing some parts of the TinyMCE views API, so if someone sees any  issues, let me know. Also, If you have any suggestions for improvements in 4.2, let me know too. :)

Thanks!

https://core.trac.wordpress.org/ticket/31412

ellatrix:
The patch as it currently is will break things.

danielbachhuber:
@iseulde Can we package a backwards compat shim somehow then?

ellatrix:
Certainly. I'll see what I can do.

mattheu:
I've managed to update this to work fairly easily. Only concern is VIP might deploy this change before 4.2... Quickstart at least will update to trunk right?

mattheu:
Closing this issue. Master is now compatable - and there are separate issues for any outstanding problems.

PR #203, opened 2015-02-26
use setIframes

### LINKED ISSUE #202: Editing a shortcode with iframed preview duplicates the shortcode
See https://github.com/fusioneng/Shortcake/pull/200#issuecomment-76078140

It appears that the second shortcode always gets added to the beginning of the editor and, if you edit it, changing the duplicate shortcode won't cause a third shortcode to appear.

Previously #195

mattheu:
Fixes #202 
- Setting `overlay: true` on the view is a workaround for making the iframes focusable/selectable.
- If we make use of  `setIframes` instead of trying to do it ourselves, it fixes the duplicate preview bug.
- We still need a workaround for loading styles -  `getEditorStyles` although I've opened a trac ticket that will remove this requirement*. 

*I think its still relevant following the 4.2 improvements - see #181

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -369,6 +369,8 @@ var shortcodeViewConstructor = {
 
 	View: {
 
+		overlay: true,
+
 		initialize: function( options ) {
 			this.shortcode = this.getShortcode( options );
 			this.fetch();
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Where did `setIframes` come from?

danielbachhuber:
> Setting overlay: true on the view is a workaround for making the iframes focusable/selectable.

This means the media isn't actually playable though, right?

mattheu:
This is what the AV shortcode previews use. Not sure why we weren't using it TBH. 

It doesn't stop it actually being usable, because as soon as its selected, the overlay is then hidden.

More importantly it adds the controls correctly rather than us having to hack it and have it visable all the time.

danielbachhuber:
Cool. Just played around with it and it feels pretty solid. Nice work.

PR #204, opened 2015-02-26
Only show inner content field when shortcode supports it 

danielbachhuber:
See #93


danielbachhuber:
cc @jitendraharpalani58

Collizo4sky:
Thanks for this update :+1:

PR #209, opened 2015-02-27
4.2 compatability

mattheu:
Fixes changes to TinyMCE views in 4.2 - re https://github.com/fusioneng/Shortcake/issues/181

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -437,6 +437,11 @@ var shortcodeViewConstructor = {
 	 */
 	edit : function( shortcodeString ) {
 
+		// Backwards compatability for WP pre-4.2
+		if ( 'object' === typeof( shortcodeString ) ) {
+			shortcodeString = decodeURIComponent( $(shortcodeString).attr('data-wpview-text') );
+		}
+
 		var model, attr;
 
 		var megaRegex = /\[([^\s\]]+)([^\]]+)?\]([^\[]*)?(\[\/(\S+?)\])?/;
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Can we keep 4.1 code in for compat?

mattheu:
Yeah - if we keep the old `View` property around - its fairly easy to support both simultaneously. Its a bit horrible though.

danielbachhuber:
Let's do that for now

danielbachhuber:
This seems to work pretty well, except it breaks Fusion Gallery preview for 4.2

danielbachhuber:
I think the JavaScript is being stripped out

mattheu:
> It breaks Fusion Gallery preview for 4.2

Fusion gallery was done first - the way I got the JS to initialize wasn't as neat as how you did if later - using the action to load the scripts. I have a PR ready to update this.

danielbachhuber:
> I have a PR ready to update this.

Ok. If you can do on Monday, I can push to prod then.

mattheu:
Current issues with this - that we would need to resolve before 4.2 drops, but I'd rather try and contribute fixes to core than build workarounds for this just yet.
- Styles aren't imported into the iframes automatically, and I'd rather not have to duplicate this functionaliyt from core. 
  - I've opened a ticket on trac - https://core.trac.wordpress.org/ticket/31464
  - Currently no styles are imported so some things look broken
- Sometimes iframe previews are not correctly resized to fit images. 
  - created a trac ticket for this also - https://core.trac.wordpress.org/ticket/31480

PR #212, opened 2015-03-01
144 Custom Attributes

### LINKED ISSUE #144: Form enhancements
I'd love to see – in the Post Element Details Edit tab – a richer form experience. 

Field descriptions (below the labels) would be very handy, and maybe the ability to increase field widths based on a selection of options (small, medium, large?).

If these abilities are already in there – awesome – I'm just having trouble finding some decent docs on possibilities keys in the `attrs` array when registering a shortcode UI.

bfintal:
Fixes #144

This is a test on how to implement custom attributes in the fields #144. Not quite sure whether this is the ideal implementation with the templates, but this was the most elegant approach I could come up with as of the moment.

I added `{{{ data.custom }}}` on each field template. This is triple brackets so that we can dynamically fill it up with attribute keys that are not defined in the `defaults` array for the attribute. The values are escaped via `_.escape` before being added to make it secure.

So for example if you do:

```
'type' => 'text',
'class' => 'regular-text',
'style' => 'background: red'
```

The `class` & `style` attributes are appended to the `input` html:

`<input type="text" name="src" id="src" value="" placeholder="" class="regular-text" style="background: red">`

This functionality is helpful for adding `multiple` in select attributes, changing input widths, adding `min`, `max`, `step`, `rows`, `cols` attributes, etc.

One caveat is that all the _normal_ attributes should be defined in the `default` list or else it would be included inside the rendered template.


danielbachhuber:
@bfintal Can you merge master?

inc/templates/edit-form.tpl.php
```diff
@@ -18,7 +18,7 @@
 <script type="text/html" id="tmpl-shortcode-ui-field-text">
 	<div class="field-block">
 		<label for="{{ data.attr }}">{{ data.label }}</label>
-		<input type="text" name="{{ data.attr }}" id="{{ data.attr }}" value="{{ data.value }}" placeholder="{{ data.placeholder }}"/>
+		<input type="text" name="{{ data.attr }}" id="{{ data.attr }}" value="{{ data.value }}" placeholder="{{ data.placeholder }}" {{{ data.custom }}}/>
```
danielbachhuber, line 21: Can we call this `data.attributes`? `custom` isn't very clear, imo.

Also, `placeholder` should be migrated to this new pattern, and we should have some backwards compat shim.

danielbachhuber:
Apologies for the late review. This looks pretty good. Just a couple minor nits.

PR #213, opened 2015-03-02
use global variable to create new instances

jitendraharpalani58:
@danielbachhuber 

to allow `extend`, we will need to use global `sui` variable to create new instances of views, controllers etc. This fix did work for select2 UI. 

I am thinking that users will want to extend other views also in future. It would make sense to make this change for the all the instances. 

Make sense?


danielbachhuber:
:+1:

danielbachhuber:
Previously #162

PR #218, opened 2015-03-04
Optimize js

mattheu:
This PR looks like a lot of changes... but its not really.

A couple of issues here:

I'm a little unsure about the way we were compiling the JS using browserify.

From Gruntfile.js `'js/build/shortcode-ui.js' : ['js/**/*.js', '!js/build/**', '!js/field-attachment.js', '!js/field-color.js']`

This just compiles all the JS files we have into one big file. 
- Unused modules such as `js/utils/dom.js` are included
- The field JS is loaded separately anyway - so this is causing duplicate code.

I think we should be compiling only the files we want and as long as we correctly require modules, browserify will do the rest.

Like so:  `'js/build/shortcode-ui.js' : ['js/shortcode-ui.js']`

Secondly.

Before we split up the JS, we were heavily using the global ShortcodeUI to provide easy reference to models/collections/views etc. Browserify mostly makes this redundant and we can remove _almost_ all use of it. 
- Global is still neccessary where we want plugins to be able to extend/modify things. A reference to all the available shortcodes for example. 
- Currently fields written as plugins use the global sui.view to register their custom views. I've kept sui.views around for this. It can probably be improved though, but we can tackle this separately.


danielbachhuber:
Looks good. Can you merge master though, and add some basic test coverage?

mattheu:
Added JS tests. 

There's pretty good coverage for all models and the MCE View constructor.

I've made a few smaller changes to make things more testable.

danielbachhuber:
Hawt! I'll do a final review of this today.

PR #225, opened 2015-03-10
Optimize JS

danielbachhuber:
Originally #218

Reverts the revert fusioneng/Shortcake#224


danielbachhuber:
@mattheu Some issues I ran into with this on Fusion:
- `[fusion-gallery]` didn't render a preview at all.
- When I tried to insert a new gallery, the `inner_content` textarea appeared when it shouldn't have.
- `[fusion-facebook-post]` had a buggy preview in that I got the preview loading indicator but it didn't make an AJAX request.

Sorry for not catching this before. I must've had browser cache or something.

PR #232, opened 2015-03-18
Updated JS to respect jQuery.noConflict() mode in WP

roborourke:
Currently the plugin is forcing the `jQuery` variable to be set back to `$`. This can have some odd interactions with other plugins in some cases. I discovered this when another plugin in use (multilingual-press-pro) calls `jQuery.noConflict()` again for some reason. That reset the `$` var causing shortcake to error.


danielbachhuber:
Ooh, I'm going to ship a v0.2.1 for this. Seems like it'll be a nasty bug.

PR #233, opened 2015-03-18
Fix noconflict

mattheu:
cc @sanchothefat @danielbachhuber.

https://github.com/fusioneng/Shortcake/pull/232 does fix the problem and I don't think we need change this with any urgency. However I don't think it fixes the real issue.

The problem is not that we're using the global, either `$` or `jQuery`. The correct solution is to properly require jQuery and assign it to $. 

The copy of jQuery used in the tests calls `jQuery.noConflict()`. Just need more test coverage!


danielbachhuber:
> Just need more test coverage!

Always the answer to everything :)

roborourke:
@mattheu the copy of jquery shipped with WP is the same and has the noConflict call too. Will have a look at how to test for it properly

mattheu:
Fixed this up so it merges cleanly. This should all be good.

danielbachhuber:
Looks good

PR #245, opened 2015-03-27
Attributes without quotes and extra template options

toolstack:
- Attributes without quotation marks are parsed correctly.
- New template options to control the size, rows, cols of the input fields.
- New template option for required fields to add an asterisk after the label (should probably add some javascript to block the update call without having values).

### DIFF
--- a/inc/templates/edit-form.tpl.php
+++ b/inc/templates/edit-form.tpl.php
@@ -17,8 +17,8 @@
 
 <script type="text/html" id="tmpl-shortcode-ui-field-text">
 	<div class="field-block">
-		<label for="{{ data.attr }}">{{ data.label }}</label>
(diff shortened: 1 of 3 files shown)


bfintal:
As discussed in #144, instead of handling specifically these additional attributes, we should just handle all attributes as in PR #212 

+1 for unquoted attributes

danielbachhuber:
> Attributes without quotation marks are parsed correctly.

Can you break this into a separate PR please, and make sure to add test coverage?

@bfintal Apologies for the tardiness on your PRs. I'm back to 100% next week, and I'll dive into them then.

toolstack:
I can break it out in to separate PR's, but I'm not familiar with building test coverage.

danielbachhuber:
Continued in #249

PR #238, opened 2015-03-22
Set defaults for inner_content

jitendraharpalani58:
See #237

PR #246, opened 2015-03-28
Inner content always shown.

mattheu:
Bug introduced here. Fixes #238. 

By setting the default values of the inner content model, it caused it always to be shown.
- I think setting the defaults is on InnerContent model is OK.
- But innerContent should not be set by default on the Shortcode model.
- Also need to update a few places to handle this not being set.
- Also tests need updating to handle this.

I've also removed the setting of defaults in PHP - as this duplicates functionality already handled by JS.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -34,18 +34,8 @@ class Shortcode_UI {
 
 	public function register_shortcode_ui( $shortcode_tag, $args = array() ) {
 
-		$defaults = array(
-			'label'         => '',
-			'attrs'         => array(),
-			'listItemImage' => '',   // src or 'dashicons-' - used in insert list.
-			'inner_content' => false,
-		);
-
-		$args = wp_parse_args( $args, $defaults );
-
-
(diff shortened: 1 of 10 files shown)


mattheu:
Review & Merge

inc/class-shortcode-ui.php
```diff
@@ -75,13 +65,6 @@ public function register_shortcode_ui( $shortcode_tag, $args = array() ) {
 			array_splice( $args['attrs'], $index, 1 );
 		}
 
-		// strip invalid
-		foreach ( $args as $key => $value ) {
-			if ( ! array_key_exists( $key, $defaults ) ) {
-				unset( $args[ $key ] );
-			}
-		}
-
```
danielbachhuber, line 84: Why is this being removed?

danielbachhuber:
Looks pretty good — just one question.

inc/class-shortcode-ui.php
mattheu, line 84: Probably could be tackled as a separate issue. But we're setting the defaults in the models themselves. I don't think we should replicate this in PHP as well.

PR #248, opened 2015-03-30
Template options

toolstack:
Split from previous PR as requested.

This PR adds additional options to the template for size, rows, cols to options for template fields as well as a "required" option to add a \* after the label.


danielbachhuber:
We support arbitrary attributes with the `meta` argument now.

PR #255, opened 2015-04-01
Difference when I run `grunt`

danielbachhuber:


danielbachhuber:
@mattheu What's going on here?

mattheu:
I think they must have been missed from https://github.com/fusioneng/Shortcake/pull/246 - that code is present in the source.

Issue #257, opened 2015-04-01
https://core.trac.wordpress.org/ticket/31480


mattheu:
Keep an eye on this - https://core.trac.wordpress.org/ticket/31480

Arguments passed to render are changing - will need to make a small change when this gets merged.

danielbachhuber:
Will we need to shim for 4.1?

mattheu:
No - We already have backwards compat by keeping around the old view object - as the structure of this code was changed.

danielbachhuber:
@mattheu Can this be closed, or are there other changes we'll be waiting for?

mattheu:
Nope we can close.

PR #262, opened 2015-04-06
Update args passed to MCE view render

mattheu:
https://github.com/fusioneng/Shortcake/issues/257

The args have changed. Render now accepts content (optional) as the first param, and force as the second. However I don't think that `force = true` is actually required.


mattheu:
review & merge

danielbachhuber:
:+1:

PR #266, opened 2015-04-07
Remove backwards compat shim for `content`

danielbachhuber:
This will let `content become an argument in its own right


danielbachhuber:
@mattheu Can you sanity check this?

mattheu:
:+1:

PR #273, opened 2015-04-15
Make the dev plugin activatable on its own

### LINKED ISSUE #264: dev.php - Example shortcode.
We currently have dev.php that includes an example shortcode. Its great for pople getting started, as well as development.

The file just needs to be included. 
- It was brougt up in the chat this month that this could be enabled in the beta tester plugin.
- I keep committing this include by mistake ;)
  - daniel suggested a git pre commit hook.

Probably some improvements that can be made here.

danielbachhuber:
Fixes #264

### DIFF
--- a/dev.php
+++ b/dev.php
@@ -1,4 +1,23 @@
 <?php
+/**
+ * Plugin Name: Shortcode UI Example
+ * Version: v1.0
+ * Description: Adds [shortcake_dev] example shortcode to see Shortcode UI in action
+ * Author: Fusion Engineering and community
+ * Author URI: http://next.fusion.net/tag/shortcode-ui/
+ * Text Domain: shortcode-ui
+ * License: GPL v2 or later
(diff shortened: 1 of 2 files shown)


danielbachhuber:
image

danielbachhuber:
@mattheu #reviewmerge

mattheu:
Not necessarily part of this ticket - so happy to merge. But do you think we should exclude this from the plugin available on WordPress.org?

danielbachhuber:
> But do you think we should exclude this from the plugin available on WordPress.org?

Eh. Probably makes sense. But, if someone is already manually loading it, then we'll cause a fatal on upgrade. I don't suspect the risk is huge.

mattheu:
Just a thought. Currently its a developer focused tool - people need to roll their own shortcodes and the example will be a huge help.

Issue #177, opened 2015-02-19
Personalize "Insert Post Element" and "Post Element Details"


danielbachhuber:
Because we have the label for the shortcode, we can make this "Insert [Shortcode Label]" and "[Shortcode Label] Details"

danielbachhuber:
Modifying "Insert Post Element" is somewhat blocked by https://github.com/fusioneng/Shortcake/issues/222#issuecomment-93582258

danielbachhuber:
I'm fine with just #275 for now.

PR #275, opened 2015-04-15
Localize title when editing shortcode attributes

danielbachhuber:
See #177


danielbachhuber:
@mattheu what do you think of this implementation approach?

### Before

image

### After

image

mattheu:
:+1:

Issue #187, opened 2015-02-24
Easy for shortcake field input IDs to conflict


mattheu:
I was testing the checkbox PR and #182 and noticed that if I used the attribute name 'checkbox', then clicking the label didn't focus the checkbox.

This is because the ID conflicts with another field on the page.

A simple fix would be to prefix `id` and `attr` to avoid conflicts. I don't think this will have any other effects.

danielbachhuber:
Let's fix this for v0.2.0 — it would be nice to get out of the way.

PR #278, opened 2015-04-20
Better unique ID

mattheu:
See https://github.com/fusioneng/Shortcake/issues/187

If the ID isn't unique - clicking the inputs label won't always focus it. This could also cause other issues further down the line. 

The problem is that using the attribute name alone isn't a very good way to ensure uniqueness.
- Pass an ID property. 
- As a default - set the ID to `shortcode-ui-%attr%-%cid%`

This should ensure that each field view object will have a unique ID.


mattheu:
Review & merge

Issue #123, opened 2015-02-01
Single or multiple post selection list


bobbingwide:
Shortcodes such as [gallery] include parameters which require the user to specify the ID or IDs of the posts to be displayed, included or excluded. 

The logic should allow for parameters which allow multiple values to be specified. 

I have many shortcodes which also support the entry of post IDs. 

It would be nice to have native support for selecting the posts from existing content. 

This requirement may involve interaction with other parameter values that the user chooses: post_type, post_status, etcetera, but it's more likely to be controlled by hidden values.

e.g. For the gallery shortcode the IDs to be displayed in a selection list would be limited to attachments.  

For a shortcode such as [bw_pdf] the selection would be filtered with post_mime_type=attachment/pdf.

danielbachhuber:
> It would be nice to have native support for selecting the posts from existing content.

We actually have an internal Select2 field for selecting one post for a shortcode. I can understand and agree with the appeal of being able to pick multiple posts. We'll keep this in mind as an enhancement.

danielbachhuber:
@bobbingwide we've just added a `post_select` field which permits the selection of a single image. However, galleries aren't easily accommodated. Is it worth opening another issue for media selection?

mattheu:
We should be able to support galleries as part of the attachment field I think. I think we should open a ticket. I've had this request before.

Also related is #244

bobbingwide:
@danielbachhuber OK. I'll have a look and see if I can use it. 
But it's not just attachments are referenced by post ID.

mattheu:
The post select allows you to specify any query you want so should be able to select any post type.

bastienho:
Hi, the same feature for taxonomies or custom lists would be very useful

misfist:
+1 I need multi-select for taxonomies and other lists.

PR #279, opened 2015-04-20
Include post select field

mattheu:
https://github.com/fusioneng/Shortcake/issues/123

This is pretty solid - we've been using it on Fusion.

My one concern is that we're going to have to bundle Select2 into shortcake. 
- Do we have a problem with this?
- Which version should we use? Currently on `3.*` so could be updated to 4.0.
- Can we avoid conflicts with other plugins using it?
- Is it still on the cards to include this in core?


inc/fields/class-field-post-select.php
```diff
@@ -0,0 +1,167 @@
+<?php
+
+class Shortcake_Field_Post_Select {
```
danielbachhuber: This should technically be `Shortcode_UI`. All of our code-level references should be `shortcode-ui` for consistency.

inc/fields/class-field-post-select.php
```diff
@@ -0,0 +1,167 @@
...
+			self::$instance->setup_actions();
+		}
+		return self::$instance;
+	}
+
+	private function setup_actions() {
+
+		add_filter( 'shortcode_ui_fields', array( $this, 'filter_shortcake_fields' ) );
+
+		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ), 100 );
```
danielbachhuber: Needs to respect #280

inc/fields/class-field-post-select.php
```diff
@@ -0,0 +1,167 @@
...
+		$nonce               = isset( $_GET['nonce'] ) ? sanitize_text_field( $_GET['nonce'] ) : null;
+		$requested_shortcode = isset( $_GET['shortcode'] ) ? sanitize_text_field( $_GET['shortcode'] ) : null;
+		$requested_attr      = isset( $_GET['attr'] ) ? sanitize_text_field( $_GET['attr'] ) : null;
+		$response            = array( 'posts' => array(), 'found_posts'=> 0, 'posts_per_page' => 0 );
+
+		$shortcodes = Shortcode_UI::get_instance()->get_shortcodes();
+
+		if ( ! wp_verify_nonce( $nonce, 'shortcake_field_post_select' ) ) {
+			wp_send_json_error( $response );
+		}
```
danielbachhuber, line 113: Do we need a capability check in here too?

js/field-post-select.js
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+	} );
+
+	/**
+	 * Extending SUI Media Controller to hide Select2 UI Drop-Down when menu
+	 * changes in Meida modal
+	 * 1. going back/forth between different shortcakes (refresh)
+	 * 2. changing the menu in left column (deactivate)
+	 * 3. @TODO closing the modal.
```
danielbachhuber, line 167: Do we need to address this?

shortcode-ui.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * Plugin Name: Shortcode UI
- * Version: v0.3-alpha
+ * Version: v0.2.0-dev
```
danielbachhuber: Can you fix this change?

danielbachhuber:
> Do we have a problem with this?

Seems like its utility outweighs the cost. Should it be abstracted to use other types too, like Users and Terms?

> Which version should we use? Currently on 3.\* so could be updated to 4.0.

Not sure of the nuances of one vs another?

> Can we avoid conflicts with other plugins using it?

Not without namespacing it, me thinks. We should enqueue as `select2` and respect when it's already been loaded.

mattheu:
V3 and V4 are not compatable with each other. So lets keep it at version 3. Its done and other people are more likeley to be using this version too. (Including us ;))

danielbachhuber:
> V3 and V4 are not compatable with each other. So lets keep it at version 3. Its done and other people are more likeley to be using this version too. (Including us ;))

Sounds good.

danielbachhuber:
Let me know when this is ready to :ship:

mattheu:
I've removed the code that attempts to destroy the field when you switch or close whilst the field is still open. It just doesn't seem to work.  However I don't think this is neccessarily a blocker - still a useful field and we can spin this off into another ticket

@jitendraharpalani58  I think you wrote this originally? Maybe you can shed some light on it?

danielbachhuber:
> I've removed the code that attempts to destroy the field when you switch or close whilst the field is still open. It just doesn't seem to work

I've reverted. It works in the insert context, but not the edit context.

inc/fields/class-field-post-select.php
danielbachhuber: 1146f8a03d488b8b1988e815b9308b79879e1861

inc/fields/class-field-post-select.php
danielbachhuber, line 113: It looks like `wp_ajax_find_posts()` doesn't have a capability check, so we're good.

PR #249, opened 2015-03-30
Unquoted attrs

toolstack:
Split from previous PR as requested.

Support unquoted values in short codes.

PR #282, opened 2015-04-21
Support for unquoted attributes

### LINKED ISSUE #241: Attribute values without quotation marks are ignored
A shortcode like:

`[testcode attr=yes]`

fails to parse the value of the attribute as it is not enclosed in quotation marks (ie. `[testcode attr="yes"]`).

The problem is on line 158 of shortcode-view-constructor.js:

`attributeMatches = matches[2].match( /(\S+?=".*?")/g ) || [];`

".*?" restricts the match to only attributes with quotation marks around their values.

mattheu:
Updated PR from toolstack - #249
- Added tests for unquoted attributes.
- Tests were failing attribute names that have a dash in them.
- Simplified regex and fixed failing tests.

This should be good to go. Thanks @toolstack

Notes 
- unquoted attributes CANNOT have spaces in them. This is consistent with core anyway.
- Unquoted attributes are converted to quoted attributes when you update them. I think this is reasonable. 

Fixes #241

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -96,8 +96,9 @@ class Shortcode_UI {
 		wp_enqueue_script( 'shortcode-ui', $this->plugin_url . 'js/build/shortcode-ui.js', array( 'jquery', 'backbone', 'mce-view' ), $this->plugin_version );
 		wp_enqueue_style( 'shortcode-ui', $this->plugin_url . 'css/shortcode-ui.css', array(), $this->plugin_version );
 		wp_localize_script( 'shortcode-ui', ' shortcodeUIData', array(
-			'shortcodes' => $shortcodes,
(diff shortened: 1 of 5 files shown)


toolstack:
Just a note that the regex I used was pulled directly from core, so anything it matches is supported in core.

mattheu:
Ah - maybe core doesn't support dashes in shortcodes then. I'll test this now

mattheu:
> Just a note that the regex I used was pulled directly from core, so anything it matches is supported in core.

OK - we had a test that passes for dashes in attribute names - but you're right that this isn't supported in core.

I think its a good call to use the regex from core. Our test for this shouldn't pass. 

I've updated my PR

mattheu:
Should be good to merge

danielbachhuber:
:+1: looks good.

PR #288, opened 2015-04-21
Reset shortcode controller when user clicks on "Insert Post Element"

### LINKED ISSUE #222: Clicking "Insert Post Element" should always take you back to the list
image

This always gets me.

danielbachhuber:
This effectively works as a back button.

Fixes #222

### DIFF
--- a/js/views/media-frame.js
+++ b/js/views/media-frame.js
@@ -54,6 +54,7 @@ var mediaFrame = postMediaFrame.extend( {
 	resetMediaController: function( event ) {
 		if ( $( event.target ).hasClass( 'media-menu-item' ) ) {
 			this.mediaController.reset();
+			this.contentRender( 'shortcode-ui', 'insert' );
 		}
 	},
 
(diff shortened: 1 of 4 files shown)


js/views/media-frame.js
```diff
@@ -44,6 +45,19 @@ var mediaFrame = postMediaFrame.extend( {
...
+		return _.extend( {}, postMediaFrame.prototype.events, {
+			'click'    : 'resetMediaController',
+		} );
+	},
+
+	resetMediaController: function( event ) {
+		if ( $( event.target ).hasClass( 'media-menu-item' ) ) {
+			this.mediaController.reset();
+			this.contentRender( 'shortcode-ui', 'insert' );
+		}
```
danielbachhuber, line 58: @mattheu Can you think of a more glamorous way to do this?

danielbachhuber, line 58: Figured it out f925bbf9b628b2e710e1bf103bd29b7244b5034e 033d398d512a3bb187e77053a1c6a7d4fe99715a

danielbachhuber:
@mattheu #review?

PR #291, opened 2015-04-21
Custom HTML attributes for fields

mattheu:
This was almost there - but I'm not sure it might be a little bit 'magic'.

I think it would be better to explicitly set custom attributes, rather than using the diff between  the attributes and the defaults.

What i've done here:
- I've updated to reflect this change.
- Handle boolean attributes
- Treat placeholder as custom attribute
- Support legacy placeholder
- Fix tests to reflect new data structure


mattheu:
review & merge

dev.php
```diff
@@ -103,9 +103,12 @@
 					'label' => 'Cite',
 					'attr'  => 'source',
 					'type'  => 'text',
-					'placeholder' => 'Test placeholder',
+					'customAttributes' => array(
```
danielbachhuber: Could we call this `meta`, inline with nav menus?

js/build/field-attachment.js
```diff
@@ -213,8 +213,21 @@ var ShortcodeAttribute = Backbone.Model.extend({
...
+		description: '',
+		customAttributes: {
+			placeholder: '',
+		}
 	},
+
+	initialize: function() {
+
+		// Handle legacy custom attributes.
+		if ( this.get('placeholder' ) ) {
```
danielbachhuber: Can we have this compatibility shim just defined in one place, and file an issue to remove it in v0.4?

danielbachhuber:
@mattheu seems fine. just a couple comments to address.

dev.php
danielbachhuber: I'd also be open to calling it `attributes` or `field_attributes` if we need to be specific.

mattheu: Renamed to meta. I think that an attributes property is already used by backbone so don't want to cause confusion.

js/build/field-attachment.js
mattheu: > defined in one place

Not sure what you mean here. I've moved alongside the other code that handles these custom attributes. 

Ticket for removal: #292

mattheu:
Shold be good to go

js/build/field-attachment.js
```diff
@@ -361,6 +364,35 @@ var editAttributeField = Backbone.View.extend( {
...
+
+			} else {
+
+				_meta.push( _.escape( key ) + '="' + _.escape( data.meta[ key ] ) + '"' );
+
+			}
+
+		}
+
+		data.meta = _meta.join( ' ' );
```
danielbachhuber, line 394: It would be nice to abstract this out #293

Issue #301, opened 2015-04-24
Continuously fetching shortcode preview bug when Shortcode content is empty.


mattheu:
If a shortcode preview response is successful but empty, then shortcake will attempt to fetch the preview again. This will happen continuously.

The fetch is triggered in the `getContent` method - when no content is found. 

However a super-simple fix for this is to trigger the ajax request on initialization. 

This will fix the infinite requests bug, but the preview will still be broken - MCE views will continue to show the preview as loading if content is empty. But I think this should be up to the shortcode author to provide some sort of feedback?

danielbachhuber:
I think an error message would be appropriate. Or our own view that says "this post element has no output"

For all the shortcodes we have, I haven't yet started to throw misconfiguration notices for when the shortcode has no output.

mattheu:
Actually - I think typically a shortcode shouldn't display an error notice on the front end of a site. Failing silently would often be better.

I'm up for having some kind of notice.

mattheu:
Fixed in #302.

Opening a new issue to discuss error mesages. #303

PR #302, opened 2015-04-24
Trigger fetch preview on initialize, not getContent

mattheu:
#301

Removed getContent method - no need provide our own version of this method. Also removed tests for getContent.

Backported fix to 4.1 compatability shim


inc/class-shortcode-ui.php
```diff
@@ -64,7 +64,11 @@ public function get_shortcode( $shortcode_tag ) {
 
 	}
 
-	public function action_after_setup_theme() {
+	/**
+	 * Action admin scripts.
+	 */
+	public function action_admin_enqueue_scripts() {
```
danielbachhuber, line 70: Why did you change this action?

mattheu, line 70: This wasn't working at all. `after_setup_theme` must be too early.

mattheu:
editor styles CSS file wasn't loading at all. So I removed a bunch of they styles since we don't seem to need them!

Kept styles for error messages.

When shortcode preview is empty - just display the formatted shortcode.

I wonder if we should do this for errors too? In that case we can simply return a JSON error if the shortcode content is empty.

danielbachhuber:
> When shortcode preview is empty - just display the formatted shortcode.

Screen grab?

> I wonder if we should do this for errors too? In that case we can simply return a JSON error if the shortcode content is empty.

Could. I don't have a strong preference. Your call.

mattheu:
image

I think we should keep the error message as is. If you're seeing this - theres a problem ;)

PR #307, opened 2015-04-27
Force MCE view render

### LINKED ISSUE #300: Creating 2 same shortcodes turn the first into its loading state
Replicate:
- Create a new page
- Create an X shortcode (any)
- Create another X shortcode (the same)

Result: result

Tested with a master copy a few days ago, version v0.3-alpha. And testing in WP 4.2-RC4-32277

mattheu:
Fixes #300 

This bug was introduced when we added compatability for the changes to MCE views introduced in WP 4.2

Previously - we called `render( true ) // force`

However in 4.2, the first param is the content, the second param is force (bool). This was changed. 

This plugin reverts this change - the behaviour should be the same as 0.2 on WP 4.1


danielbachhuber:
Travis is broken again :(

PR #314, opened 2015-04-28
Allow shortcode ui to register default values for attrs

goldenapples:
Adds an additional field to shortcode attribute data model called 'default_value'. If a default_value is set and the value is empty, the value will be set to the default value.

Can I get a review of this? @jitendraharpalani58 @noppanit @danielbachhuber 
This attribute would be useful to me for https://github.com/fusioneng/image-shortcake and I'd like to have it added here if possible. This is the first time I've looked at Shortcake internals and I don't know if this is the best way to do this, though...

### DIFF
--- a/js/src/views/edit-attribute-field.js
+++ b/js/src/views/edit-attribute-field.js
@@ -31,6 +31,11 @@ var editAttributeField = Backbone.View.extend( {
 			delete data.placeholder;
 		}
 
+		// Apply default value if one exists and no value is set yet
+		if ( '' === data.value && data.default_value.length > 0 ) {
+			data.value = data.default_value;
+		}
+
 		// Convert meta JSON to attribute string.
 		var _meta = [];
 		for ( var key in data.meta ) {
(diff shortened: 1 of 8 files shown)


js/src/views/edit-attribute-field.js
```diff
@@ -31,6 +31,11 @@ var editAttributeField = Backbone.View.extend( {
 			delete data.placeholder;
 		}
 
+		// Apply default value if one exists and no value is set yet
+		if ( '' === data.value && data.default_value.length > 0 ) {
```
danielbachhuber, line 35: How would you set an empty value?

goldenapples, line 35: Good question. Probably this needs more thought put into the data structure to hold it.

I was only thinking of the case of selects / radio buttons, where it would be easy to make a value required. Text fields where an empty string is a valid choice, but a default value is still useful, add an extra complication.

goldenapples:
Closing this. I was misunderstanding the role of the `value` field, which does something like what I wanted to do here. A better approach is something more like #323 - treating the first value in a select or radio button list as the `value` if none is otherwise set.

Issue #317, opened 2015-05-01
UI Breaks inline shortcodes. Adds line breaks.


brentjett:
The UI treats all registered shortcodes as block level elements and adds line breaks when they are inserted into a sentence. Some shortcodes are meant to be inline and are broken by the visual editor. Example:

> Call us at [main_phone] to get started.

This shortcode would be broken into multiple lines by the UI plugin and then saved as:

> Call us at
> 
> [main_phone]
> 
> to get started.

danielbachhuber:
> Some shortcodes are meant to be inline and are broken by the visual editor

I'm filing this as an enhancement, because we've always treated previews as blocks, but inline seems like a reasonable enhancement.

brentjett:
Ok thanks! For the moment it seems the best way to keep these inline shortcodes from breaking is to not register them with shortcake. Would be great to be able to insert them from the UI like the others. Thanks! Great project. I might be interested in helping.

bobbingwide:
+1

bobbingwide:
This additional comment may actually turn out to be the opposite. 
I have a shortcode which expects to see line breaks in the content.
e.g. `[bw_csv]` Use to display the contents of the shortcode in a tabular form.

```
[bw_csv]uo= parameter,Format as
default,Table
uo=u,Unordered list
uo=o,Ordered list
uo=d,Definition list
[/bw_csv]
```
- The heading row immediately follows the shortcode
- Every other row of the table is on a separate line.
- Blank lines are ignored.

When using `Insert Post Element > Preview` the table displays correctly
After `Insert Element` the new lines are lost.

More often than not Preview in the Visual editor doesn't expand the shortcode. 
It appears that do_shortcode is not happy with the `<br />` in the content.
Note: This could be a local environment issue.

In other cases the original content from the Text editor is ruined when switching to Visual and then back again. Obviously I have some more problem determination to perform here!

Suffice it to say shortcake doesn't  handle this at all well:

```
[bw_csv]Shortcode,Inner content
[bw_code bw_csv],[bw_dash yes]
[bw_code bw_geshi],[bw_dash yes]
[bw_code caption],[bw_dash yes]
[bw_code wp_caption],[bw_dash yes]
[/bw_csv]
```

See what it's supposed to come out as here
http://www.oik-plugins.com/shortcode_example/bw_csv-to-create-a-table-with-nested-shortcodes/

bobbingwide:
[...]

Issue #169, opened 2015-02-17
Support for quick edit buttons


danielbachhuber:
Core's embedded image representation has these nifty alignment quick-edit buttons:

![image](https://cloud.githubusercontent.com/assets/36432/6235133/0a95fec0-b695-11e4-89de-27bca251d4b9.png)

It would be neat if Shortcake:
- Optionally supported these alignment parameters for shortcodes
- Supporting registering arbitrary "Quick edit" shortcode buttons

soderlind:
+1

mattheu:
+:100: on the alignment

soderlind:
@mattheu agree, and with alignment being able to set the width is need.

mattheu:
The issue here however is the wp mceView doesn't really handle this at all...

soderlind:
Hopefully the work [being done](https://make.wordpress.org/core/2015/01/22/the-case-for-js-modules/) will make it easier to grok wp.*

dovy:
Clone anyone?  ;)

danielbachhuber:
Bumping this out of v0.4.0. It needs a greater degree of technical clarity before we commit it to a milestone.

cliffordp:
Could add custom ones too, like this: https://github.com/gambitph/Page-Builder-Sandwich/wiki/Adding-Toolbar-Buttons
![](https://github.com/gambitph/Page-Builder-Sandwich/wiki/images/toolbar.jpg)

danielbachhuber:
For v0.5.0, if this isn't technically viable with a reasonable amount of development effort, we should at least establish what we need to do.

bruno-barros:
wow guys. This PBS did an amazing job with shortcodes.
I dream with these two projects working together.

cliffordp:
PBS relies on Shortcake

bruno-barros:
Yes, of course, but Shortcake could bring some features from PBS like the contextual menu 
if it make sense.

ivankristianto:
I want to bump this thread. Wondering if there is any progress on this feature?

PR #320, opened 2015-05-06
Force mce view sandbox to clear floated elements inside it

goldenapples:
Shortcodes that contain floated elements break in preview because the iframe body collapses to zero height, leaving the floated element invisible.

This avoids that problem by using the `display: inline-block` / `overflow: hidden` on the body of the wpview iframe.

Addresses the problem discussed in #9.

### DIFF
--- a/css/sass/shortcode-ui-editor-styles.scss
+++ b/css/sass/shortcode-ui-editor-styles.scss
@@ -16,5 +16,10 @@
 		color: #666;
 		font-size: 14px;
 	}
+}
 
+body#wpview-iframe-sandbox {
+	display: inline-block;
+	width: 100%;
+	overflow: hidden;
 }
--- a/css/shortcode-ui-editor-styles.css
+++ b/css/shortcode-ui-editor-styles.css
@@ -8,4 +8,9 @@
   color: #666;
   font-size: 14px; }
 
+body#wpview-iframe-sandbox {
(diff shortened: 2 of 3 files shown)


goldenapples:
@danielbachhuber thoughts on this, to solve the problem of floated elements in previews being cut off?

goldenapples:
One thing that I think could be improved on is the `width: 100%`. I'm fairly sure that by making some creative use of `width: fit-content` on the body and html elements of the iframe, it should be possible to have inline or floated shortcodes, making the edit preview that much more useful. That probably needs some more work, though, to find the first element inside the iframe, check its float property, and apply that same float property to the `wpview-body` element surrounding it.

danielbachhuber:
> thoughts on this, to solve the problem of floated elements in previews being cut off?

Seems like a reasonable approach to me. @mattheu ?

> That probably needs some more work, though, to find the first element inside the iframe, check its float property, and apply that same float property to the wpview-body element surrounding it.

Related https://github.com/fusioneng/Shortcake/issues/317#issuecomment-98203636 and #169

I actually chatted with @gcorne (one of the original TinyMCE views authors) a bit today about this non-block element mess. It seems like it might be ripe for a solution :)

danielbachhuber:
This seems to work pretty well for our immediate needs.

Issue #236, opened 2015-03-19
Add rich text as a potential field


danielbachhuber:
For shortcodes where you might want to grant the user basic HTML, it would be helpful to apply TinyMCE to a field and grant basic WSYWIG access.

hirejordansmith:
Yes Please!

chrisvanpatten:
Ooooh yes this'd be great!

mattheu:
I think that this is a blocker for this one: https://github.com/fusioneng/Shortcake/pull/179

mattheu:
Pusing to 0.4.

Perhaps we could encode this data in order to work around the complications with nested HTML in shortcodes.

danielbachhuber:
We can now support this for `inner_content`

ghost:
@danielbachhuber - just taken my first look at Shortcake and think it looks great!

WYSIWYG would be great (& enable me to use Shortcake on client projects).  Is it still in the pipeline? ~~(& is there anyway to get HTML in inner_content now?) edit: already possible!~~

Thanks!

ryanboswell:
So wondering if there was someone working on this already and if it was anywhere near ready for release for the v0.5.0 milestone it's currently tagged to, I don't see any open PRs or released code for it so far. I'm considering helping out on this and submitting as a PR, but wanted to check before diving in.

danielbachhuber:
I think #322 is the latest and greatest attempt. Now that core has fixed itself, we can easily support HTML for `inner_content`. I'm still wary of committing to an encoded attribute approach for everything else though.

Can you share your use case in a bit more detail?

ryanboswell:
Ah, wasn't sure where that one stood.

The use case I'm thinking of is a shortcode that wraps rich text (aka. HTML markup) in specific markup that applies custom styling. The idea being that we want to grant some level of custom styling allowance to editors/writers, but currently don't allow them to add in-line style attributes in post content. So we accomplish that with a shortcode that hand holds them through the process and keeps them from breaking things design-wise. You can think of it kinda like a blockquote shortcode on steroids. Ideally we'd be able to allow them to edit the contents of the shortcode using a TinyMCE editor, since many editors/writers aren't all that familiar with writing HTML.

danielbachhuber:
Cool, makes sense. It seems like you'd be able to do that with `inner_content`

ryanboswell:
[...]

PR #322, opened 2015-05-07
Allow HTML in attributes

roborourke:
This PR adds support for escaping html in attributes so that HTML no longer breaks the view rendering.

It should also go some way to paving the way forward for https://github.com/fusioneng/Shortcake/issues/236

It does a few things:
1. Adds a default attribute for field types called 'escape'
2. If 'escape' is true the value is passed through `encodeURIComponent()` and `decodeURIComponent()` where appropriate
3. Shortcodes with fields that are marked to be escaped have a filter added to run `rawurldecode()` automatically on the values provided the shortcode author uses the `shortcode_atts()` function and provides the 3rd parameter stating the shortcode tag name

Example:

``` php
<?php

add_shortcode( 'html_textarea_field', function ( $attr = '', $content = false ) {

    $attr = shortcode_atts( array(
       'html_content'   => '',
    ), $attr, 'html_textarea_field' ); // 3rd param here enables the automatic rawurldecode() filter

    $output = apply_filters( 'the_content', $attr['html_content'] );

    return $output;
} );

    shortcode_ui_register_for_shortcode( 'html_textarea_field', array(
        'label'         => __( 'HTML Textarea' ),
        'listItemImage' => 'dashicons-align-left',
        'attrs'         => array(
            array(
                 'label' => __( 'HTML' ),
                 'attr'  => 'html_content',
                 'type'  => 'textarea',
            )
        )
     );

}
```

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -50,6 +50,21 @@ class Shortcode_UI {
 		$args['shortcode_tag'] = $shortcode_tag;
 		$this->shortcodes[ $shortcode_tag ] = $args;
 
+		// filter the attrs to decode fields with escape=true
+		add_filter( "shortcode_atts_{$shortcode_tag}", function( $out, $pairs, $atts ) use ( $args ) {
+
+			$fields = Shortcode_UI_Fields::get_instance()->get_fields();
+
+			foreach( $args['attrs'] as $attr ) {
+				if ( ! $fields[ $attr['type'] ]['escape'] ) {
+					continue;
+				}
+				$out[ $attr['attr'] ] = rawurldecode( $out[ $attr['attr'] ] );
+			}
+
+			return $out;
+		}, 1, 3 );
+
 	}
 
 	public function get_shortcodes() {
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -55,7 +55,7 @@ class Shortcode_UI {
 
 			$fields = Shortcode_UI_Fields::get_instance()->get_fields();
 
-			foreach( $args['attrs'] as $attr ) {
+			foreach ( $args['attrs'] as $attr ) {
 				if ( ! $fields[ $attr['type'] ]['escape'] ) {
 					continue;
 				}
--- a/inc/fields/class-shortcode-ui-fields.php
+++ b/inc/fields/class-shortcode-ui-fields.php
@@ -8,6 +8,7 @@ class Shortcode_UI_Fields {
 	private $field_defaults = array(
 		'template' => 'shortcode-ui-field-text',
 		'view'     => 'editAttributeField',
+		'escape'   => false,
 	);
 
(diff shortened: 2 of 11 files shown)


danielbachhuber:
I'm :-1: on storing escaped data in attributes. It seems like a technical hack around conceptual contradiction — shortcodes are macros to produce HTML markup.

goldenapples:
I think escaping the data in some way is unavoidable. Its a common use case for shortcode attributes to hold [names](http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/), strings for display, or other data which can't reasonably be expected to be free of all special characters. There's no other way, short of a core change, to get `shortcode_parse_atts` to handle quotes or some other special characters in attribute text. 

The idea of the filter on `"shortcode_atts_{$shortcode}"` seems like a good way to handle this transparently for those attributes that desire it, while still preserving existing default behavior. Yes, this isn't what urlencoding is really for, but it works, and has the advantage of being available in a similar form in PHP and javascript.

goldenapples:
But, if we're looking for alternative approaches, replacing the attribute parser regex in core with something like an XML attribute reader would be a good start. My initial confusion at this issue came because intuitively I expected shortcode attributes to behave like HTML attributes and was surprised that `attr="value with \"quotes\""` didn't work. We already have tools to deal with escaping attribute values; if shortcode attributes worked like XML/HTML attributes, then the problem here would be simpler and we could use a more appropriate escaping method.

roborourke:
I think the main problem I was getting around was that it breaks the view if an attribute has html in, when you have a textarea field that gets stored in an attribute this can happen with no clear way to recover easily.

URL escaping was the only common way to handle this between JS and PHP besides `atob()` / `btoa()` which aren't supported before IE10, guess that depends on the browser support offered by wp admin which afaik is still IE8.

I agree it's totally hacky but can't see how else would #236 be feasible?

roborourke:
A slightly less hacky approach may be to embed the data inside `inner_content` somehow. Eg. `<div style="display:none" data-shortcode-attr="attribute-name">some <strong>HTML</strong> content</div>` then read the values out from there

roborourke:
@mattheu any thoughts on this?

danielbachhuber:
It would be nice to have an official core opinion on this.

danielbachhuber:
We [chatted about this some today](https://wordpress.slack.com/archives/feature-shortcode/p1431372428000042). It seems like encoding behind the scenes is the only way we can go for now.

Could we make registration happen on the field-level, and automatically unencode so no modification to an existing shortcode is necessary?

danielbachhuber:
@sanchothefat One nit before this lands — could we call the argument `encode` instead of `escape`? Technically we're doing the former.

danielbachhuber:
Actually, now that #179 has landed, let's bump this out of v0.4.0 and come back to it if we need.

danielbachhuber:
Closing in favor of #496

Issue #319, opened 2015-05-06
Attachment Field: Include loading indicator while image is loading


danielbachhuber:
At the moment, while the image is loading, there's a flash of UI that looks broken. We should have some loading indicator so the user knows their attachment isn't missing.

PR #324, opened 2015-05-07
Add loading indicator for attachment selector

goldenapples:
Uses a slightly modified version of the wpview loading indicator
animation to show that an image is being loaded, when first initializing
a shortcode UI containing an attachment field.

Screenshot:

screenshot

Addresses #319.


danielbachhuber:
:100:

Issue #323, opened 2015-05-07
Require that select elements have a value set.


goldenapples:
Currently, the shortcode attribute model's `updateValue` callback only happens on `change` events. 

This is confusing with attribute fields like selects, radio buttons, and checkboxes, where the user would assume that the state the field is in when the form is initially rendered -- the first item in a select list, or an unchecked value for a checkbox -- is the value they are selecting. However, unless the user actually changes the select to a different value and then changes it back, that attribute will simply not exist on the shortcode model.

I think a more intuitive approach would be, if no `value` is set on an attribute when registering shortcode ui for a shortcode, to default the `value` to the first item in the list. This would preserve the expected behavior of HTML forms.

fourfour:
+1

danielbachhuber:
:+1:

danielbachhuber:
No longer applicable https://wordpress.slack.com/archives/feature-shortcode/p1433790569000071

cggit:
this seems to still be an issue

PR #329, opened 2015-05-19
Require select elements to have a value set

goldenapples:
On registering UI for a shortcode, checks that any attributes of type "select" have a value set. 

If not, sets their value to the key of the first option, as this is the behavior that will be expected by the user who sees that the first element in the select list appears to be selected on the form.

See #323.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -47,6 +47,13 @@ class Shortcode_UI {
 			$args['attrs'] = array();
 		}
 
+		foreach ( $args['attrs'] as &$attr ) {
+			if ( 'select' === $attr['type'] && ! isset( $attr['value'] ) ) {
+				$_shortcodes = array_keys( $attr['options'] );
+				$attr['value'] = array_shift( $_shortcodes );
+			}
+		}
+
 		$args['shortcode_tag'] = $shortcode_tag;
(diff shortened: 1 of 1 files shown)


goldenapples:
I'm really not sure where this logic should live. I tried it in a couple of different places: in the  `ShortcodeAttribute` model and in the `editAttributeField` view. 

Putting this logic in the view would make the most sense to me, as that's what it really affects, but the logic involved there was ugly, since the attribute inherits defaults which could be falsy.

An initialize function in the attribute model could work, but this was this simplest approach I could think of. As a matter of fact, I kind of think some more solid validation of shortcode attributes probably should be done in PHP before sending the shortcode attrs to javascript.

I'd like to hear thoughts on this if anyone has ideas.

danielbachhuber:
@sanchothefat Any opinions?

mattheu:
Hmm... i'm not sure this is the right place for this. Although it is nice and simple.

I think this should probably go in the view, or possibly the attribute model.

goldenapples:
> I think this should probably go in the view, or possibly the attribute model.

This was also made moot by https://github.com/fusioneng/Shortcake/pull/336, which now triggers the view's `updateValue()` method on render, setting the model to the initial state of the view. So this pull can be closed.

PR #332, opened 2015-05-19
Use "regular-text" class for inputs in WordPress admin

goldenapples:
WordPress has some distinctive admin styles for form elements. Applying the base classes which WP uses to mark up form elements will make the shortcode UI form look less incongruous with the rest of the admin area. It's also easier to enter text with a wider input (`regular-text` is 25em wide; browser defaults for input width vary, but can be as low as 150px on Firefox)

_Text boxes, before:_

screen shot 2015-05-19 at 4 05 43 pm

_After:_

screen shot 2015-05-19 at 4 07 48 pm


danielbachhuber:
:cactus:

Issue #333, opened 2015-05-21
Conditionally display a field based on value of another field


danielbachhuber:
A [request from the forums](https://wordpress.org/support/topic/once-shortcode-multiple-forms-for-each-option?replies=1):

> After clicking the ShortcodeUi's shortcode edit button in the editor, shortcodeUi should display an select list with options like "button", "CTA Button" & "Share Button".
> 
> If user selects "CTA button" then ShortcodeUI should display "CTA_id" field.
> If user selects "Share Button" then ShortcodeUI should display "Share_id" field.

Fieldmanager has some `display_if` flag we could emulate.

goldenapples:
I think that this can be covered along with #316, as I imagine a lot of the same use cases for each. Registering a js callback function with the shortcode ui that receives the new value and has access to the collection to modify it would be one way of accomodating this.

goldenapples:
The system of JS actions introduced in #336 makes this possible. Do we want a syntactic sugar for this specific use case, so that an attribute in the shortcode UI registration array can generate the javascript event listener to show/hide fields conditionally? Or is letting authors register their own js callbacks sufficient? Either way, this will need some documentation.

danielbachhuber:
> Do we want a syntactic sugar for this specific use case, so that an attribute in the shortcode UI registration array can generate the javascript event listener to show/hide fields conditionally?

Not at this time. The Fieldmanager implementation is actually a bit buggy, so I don't see value in copying it 1:1.

> Either way, this will need some documentation.

Pull request against `readme.txt`?

Issue #316, opened 2015-04-29
Implement events on attribute fields that can be listened to


goldenapples:
I'm working on an issue that requires listening to changes on an attachment field, and updating other fields in a shortcode if they're not already set with the values from the attachment.

This seems as though it would be much easier to do if each field surfaced some custom events which could be listened to by another controller.  Would it make sense for each view to trigger an event in its `updateValue` callback which could be listened to by other observers? Or is there already a way of doing this which I'm not seeing? 

More generally, I guess what I'd like is an API for extending attribute fields with custom scripts. Is this being discussed, or does anyone how ideas how best to approach it?

noppanit:
I don't think it's been discussed. Are you thinking of something like KnockoutJs or AngularJs? with observer pattern?

goldenapples:
Yeah, my initial thought was to how in AngularJs, it would be something like `$scope.$emit('eventname', data)`, then you would need to attach a listener on the rootScope to pass that data to another scope.

In Backbone, the pattern seems to be to broadcast an event from a view like `this.trigger('eventname', data)` and then listen from another backbone model like `this.listenTo( editAttributeField, 'eventname', callback )`.

Seems like this could be accomodated by triggering events at a few relevant points (at the end of `updateValue` for editAttributeField, at the end of `updateAttachment` for attachment fields, etc. I'm not sure I'm seeing the whole architecture though.

danielbachhuber:
> I'm not sure I'm seeing the whole architecture though.

When I was working on `editAttributeField` last, I was thinking it might be useful to have some base attribute model each field should extend.

danielbachhuber:
@goldenapples Are you happy enough with this to close?

goldenapples:
Yup, I think its good and should work for my purposes. I wanted this for a couple of attributes in Image Shortcake, so I'll use it there and see if it works for everything I wanted.

I'm debating whether there should also be an action that fires at the shortcode level, so that you can register a listener on all attributes to a shortcode. But I haven't thought of a use case where that would be necessary yet, so I'll leave it out until I see one.

PR #336, opened 2015-05-27
Add basic API for registering callbacks on attributes

goldenapples:
This adds a basic method for registering a callback function on an
attribute field when registering UI for a shortcode. If the attribute
`callback` is defined on a field and maps to the name of a valid
javascript function, that function will be called on rendering the
shortcode UI and on any changes to that attribute field.

The value of _this_ in the callback function will be the array of view
models representing all of the attribute input field views, and the
function receives two arguments: the changed value, and the collection
of model attributes.

See #316 and #333.

An example of a callback function that uses this API is in https://github.com/fusioneng/image-shortcake/pull/36


goldenapples:
I would love to get people's feedback on this as a potential solution to the problem of dynamically updating or showing/hiding fields based on the value of another field. I think the function signature of the callback function probably needs some thought as to possible use cases - what information would make writing callback functions easier? - but this is the general approach I think would work for this.

danielbachhuber:
> I think the function signature of the callback function probably needs some thought as to possible use cases

Seems like a lot of code to register a callback function -_- Not sure how we'd do any better though, given it's a PHP API to register some frontend functionality.

Rather than add a `callback` argument, which restricts us to only one named callback, how do you feel stealing some of the proposal for JS actions in core ([here's the latest](https://core.trac.wordpress.org/ticket/21170#comment:84))?

goldenapples:
> Rather than add a callback argument, which restricts us to only one named callback, how do you feel stealing some of the proposal for JS actions in core (here's the latest)?

I actually like that concept a lot. There's no reason to have to name a callback function in the PHP array that registers the shortcode UI. Every update could trigger a named hook, like `wp.hooks.shortcodeUi.img.linkto`, and anything registered on that hook would be run. I'll give that approach a shot.

danielbachhuber:
> Every update could trigger a named hook, like wp.hooks.shortcodeUi.img.linkto, and anything registered on that hook would be run.

Yes, this would be pretty nice.

cc @gcorne @azaozz if you have any opinions about how we experiment with this

danielbachhuber:
Some conversation with @azaozz https://wordpress.slack.com/archives/core/p1433109163000965

goldenapples:
Yeah, that pretty much jived with what I saw looking into those libraries. I used a slightly modified version of https://github.com/carldanley/WP-JS-Hooks, and registering a listener function makes a lot more sense now. Updated [the branch in Image Shortcake that uses this](https://github.com/fusioneng/image-shortcake/pull/36/files); it hasn't changed much, but it's nice that javascript functions don't have to declare themselves in PHP anymore.

I'm not sure what the best way of including this code in the plugin is, though. I ended up including the WP Hooks EventManager class in the `/lib` directory and enqueuing it before and as a dependency of `shortcode-ui.js`. I can't think of a better way right now of including it that would indicate that it's external code, and that it should be kept up-to-date with developments in that external code.

danielbachhuber:
> I'm not sure what the best way of including this code in the plugin is

What about:
1. Putting it in `lib/wp-js-hooks/wp-js-hooks.js`
2. Modifying the file so it assigns itself to `wp.shortcake.hooks` or similar (e.g. avoid conflict with more generic `wp.hooks`).
3. Enqueue it with the last updated date as the version (e.g. `2015-03-19`), so we have that as reference.

Would this address your reservations?

goldenapples:
Yeah, that's pretty much what I did here. The process is kind of a mess, but I couldn't come up with anything better. I considered forking the `wp-js-hooks` repo, modifying it as we needed, and registering it as a bower component, so that there would be a VCS trail back to the original project. But... the feature plugin project itself isn't really what I'd be interested in tracking here, so much as the core discussion. So referencing the trac ticket in that file seemed more useful.

danielbachhuber:
Cool. If you'd like to update then, we can land it.

goldenapples:
Nice! Ready for final review, I think...

inc/class-shortcode-ui.php
```diff
@@ -96,7 +96,11 @@ public function enqueue() {
 
 		usort( $shortcodes, array( $this, 'compare_shortcodes_by_label' ) );
 
-		wp_enqueue_script( 'shortcode-ui', $this->plugin_url . 'js/build/shortcode-ui.js', array( 'jquery', 'backbone', 'mce-view' ), $this->plugin_version );
+		// Load minified version of wp-js-hooks if not debugging.
+		$wp_js_hooks_file = 'wp-js-hooks' . ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '.min' : '' ) . '.js';
+
+		wp_enqueue_script( 'wp-js-hooks', $this->plugin_url . 'lib/wp-js-hooks/' . $wp_js_hooks_file, array(), '1433195267' );
```
danielbachhuber: Can we enqueue as `shortcode-ui-js-hooks` to avoid collisions?

danielbachhuber: And what is that version?

js-tests/build/specs.js
```diff
@@ -40,7 +40,8 @@ describe( "Shortcode Attribute Model", function() {
 		description: 'test description',
 		meta:  {
 			placeholder: 'test placeholder'
-		}
+		},
+		callback:    'callbackFunc'
```
danielbachhuber: Shouldn't these instances of `callback` be removed?

danielbachhuber:
Couple small questions.

inc/class-shortcode-ui.php
goldenapples: > Can we enqueue as shortcode-ui-js-hooks to avoid collisions?

Yeah, makes more sense. I'll update.

> And what is that version?

The datetime I added that line. I guess that doesn't clear anything up at all - the chance that someone who updated the file would also remember to update the version number here is slim to none. Maybe better to just use the plugin version number like aberything else does?

danielbachhuber: Can we just use `2015-03-19` to more clearly indicate it's a date?

goldenapples: Sure. What date does that refer to? The [patch file](https://core.trac.wordpress.org/attachment/ticket/21170/21170-2.patch) I was building off of is from `2014-03-08`.

danielbachhuber: You should use the latest from the repo: https://github.com/carldanley/WP-JS-Hooks

goldenapples: Ah, yeah, that makes sense. I've updated that, and clarified in the file header comments.

Issue #337, opened 2015-05-27
Fatal error when activating TinyMce for frontend comments


StaggerLeee:
I use latest and best code for comments visual editor from WP Edit PRO plugin.

Shortcake gives: 

Fatal error: Call to undefined function get_current_screen() in D:\EasyPHP\data\localweb\some_test_wp\wp-content\plugins\shortcode-ui\inc\class-shortcode-ui.php on line 85
Disabled Shortcake plugin and editor shows in comments and working fine as before.

Not that I use now Shortcake for anything, just want you to know.

StaggerLeee:
Why not check if visual editor is activated on frontend and use Shortcake in the same way as in backend ?
Is it possible at all ?

I see many uses here. bbPress, comments, support forums, all possible media in comments to make it easy for beginners but arranged with Shortcake to exclude security risks, very customized media modal box, customized URL field, all possible fields, etc...

danielbachhuber:
> Why not check if visual editor is activated on frontend and use Shortcake in the same way as in backend ?

Shortcake will still work on the frontend. The feature where a developer can limit a given shortcode UI to specific post types will only work on the backend until #366 is fixed.

PR #338, opened 2015-05-27
Only enqueue scripts in the admin area

goldenapples:
Quick fix for #337.

Because `get_current_screen()` breaks if not in admin. Also, since shortcodes are tied to post types, it doesn't currently make sense to load shortcode UI on the frontend, such as comments area.


danielbachhuber:
> Only enqueue scripts in the admin area

What if we just bail early on loading the entire plugin?

danielbachhuber:
Continuing with #355

Issue #344, opened 2015-05-30
Double slash in editor css path


glueckpress:
https://github.com/fusioneng/Shortcake/blob/master/inc/class-shortcode-ui.php#L71 produces a double slash before `css/…`. The line should say:

```
add_editor_style( $this->plugin_url . 'css/shortcode-ui-editor-styles.css' );
```

or 

```
add_editor_style( trailingslashit( $this->plugin_url )  . 'css/shortcode-ui-editor-styles.css' );
```

(though I don’t think the latter would be necessary).

PR #345, opened 2015-05-30
#344 Fixes double slash in file path

glueckpress:
Biggest PR ever, removes a slash. :)

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -68,7 +68,7 @@ class Shortcode_UI {
 	 */
 	public function action_admin_enqueue_scripts() {
 		// Editor styles needs to be added before wp_enqueue_editor
-		add_editor_style( $this->plugin_url . '/css/shortcode-ui-editor-styles.css' );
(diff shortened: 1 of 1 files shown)


inc/class-shortcode-ui.php
```diff
@@ -68,7 +68,7 @@ public function get_shortcode( $shortcode_tag ) {
 	 */
 	public function action_admin_enqueue_scripts() {
 		// Editor styles needs to be added before wp_enqueue_editor
-		add_editor_style( $this->plugin_url . '/css/shortcode-ui-editor-styles.css' );
+		add_editor_style( $this->plugin_url . 'css/shortcode-ui-editor-styles.css' );
```
danielbachhuber, line 71: Can you `trailingslashit()` just so we're safe with this assumption always?

glueckpress:
Ah, sry, saw this too late. Funny, I _thought_ I should `trailingslashit()`, but then thought no, they probably had a reason to not `trailingslashit()` in the first place. So I didn’t. Next time, I will. :)

PR #349, opened 2015-06-03
Make sure shortcode prop is set

mattheu:
Bug introduced here: https://github.com/fusioneng/Shortcake/commit/c575c6527165c88ffbb9c77edb552c3ef295a012

Problem - the `shortcode` property on the view was not getting set correctly. Caused the following error when trying to edit a shortcode.

`Uncaught TypeError: Cannot read property 'attributes' of undefined`

Why? In recent versions of Backbone - you cannot pass arbitrary properties to a view.  This wasn't an issue before - because we just weren't using it.

This problem only affects the `inner_content` field - because for all other fields, we set the shortcode property after instantiating - which works OK.

The fix - set properties in the same way for `inner_content` field view that they are for all other fields to ensure it is actually set.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -727,12 +727,11 @@ var EditShortcodeForm = wp.Backbone.View.extend({
 		if ( innerContent && typeof innerContent.attributes.type !== 'undefined' ) {
 
 			// add UI for inner_content
-			var view = new editAttributeField( {
-				model:     innerContent,
-				shortcode: t.model,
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks @mattheu !

danielbachhuber:
cc @goldenapples

goldenapples:
Ah, thanks for the fix @mattheu. I didn't see that issue... guess I didn't test with a shortcode containing inner content.

PR #355, opened 2015-06-08
Only call `get_current_screen()` when the function exists

### LINKED ISSUE #337: Fatal error when activating TinyMce for frontend comments

danielbachhuber:
`get_current_screen()` only exists in admin context. While Shortcode UI isn't full supported on the frontend, not fataling is the least we can do

Fixes #337

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ We've removed the compatibility shim for the magical `content` attribute. If you
 * Added Chinese translation.
 * Added French translation.
 * Added Spanish translation.
+* Bug fix: Prevent fataling when editor is loaded in the frontend context.
 * Bug fix: Color field should also support `meta` argument.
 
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@mattheu How do you feel about this?

mattheu:
Agreed that there isn't really a better way to handle this.

Issue #351, opened 2015-06-04
Attachment Field: Indicate metadata about selected attachment


danielbachhuber:
When I have an attachment selected, it would be nice to see its title, type, size, and other metadata:

![image](https://cloud.githubusercontent.com/assets/36432/7990161/e220b512-0aa3-11e5-95d1-6edc1f917dda.png)

danielbachhuber:
We should also offer a link to the attachment's "edit attachment" view, where the user can change crop, etc.

danielbachhuber:
Looking good:

![image](https://cloud.githubusercontent.com/assets/36432/9255450/4cd67f3c-419e-11e5-9abe-d260ae475ab1.png)

Issue #350, opened 2015-06-04
Attachment Field: Clicking on attachment should let me pick a different attachment


danielbachhuber:
![image](https://cloud.githubusercontent.com/assets/36432/7990138/bf76ee50-0aa3-11e5-89b0-479c4df08aae.png)

davisshaver:
@mkeppel, assigning to you for review.

danielbachhuber:
One thing I noticed: it would be nice if the library retained the selection when I opened it

danielbachhuber:
I'll handle retaining the selection in v0.6.0

PR #358, opened 2015-06-10
Rework attachment field view so that it triggers events

goldenapples:
See #316.

This is a start at rewriting the "attachment" field view as a Backbone view that extends `editAttributeField`, and therefore has access to the actions defined in there. A big change is that it <del>is no longer browserified and</del> only depends on the public variables exposed in `sui.views`. I think this is a simpler approach and provides a better model to point people to for developing additional field types.

Architecturally - still needs some organizational work. I wanted to put it out for general feedback first, though.


goldenapples:
@danielbachhuber @mattheu I'd love some feedback on this approach for registering attribute fields if either of you have a chance. 

I think all of the custom attribute fields - attachment, post select, and color picker - should probably be registered with a standalone view definition like this rather than built into unnecessarily large and complex browserified beasts. This seems like a simpler pattern to ask plugin developers to follow when developing their own UI for attribute fields.

In the short term, this makes changes to the field like #350 and #351 pretty simple, and it makes listening for events as in #316 possible.

danielbachhuber:
I don't have a strong opinion one way or another at this point. Couple pieces of context:
- @jitendraharpalani58 did the JS abstraction originally in #162
- WordPress core tried to go full-Browserify, and then ran into the problem that developers had already extended Media Library code with global scope. Now core builds a bunch of JS files into a few global scope JS files.

However, I'm also hesitant to "fix something that isn't broken". As such, it makes sense to have a clear and compelling reason for the change.

goldenapples:
> a clear and compelling reason for the change

My impetus for this pull request is that 
- issues like https://github.com/fusioneng/image-shortcake/issues/12 and https://github.com/fusioneng/image-shortcake/issues/35 are hard to tackle when there isn't a way to listen for updates on the attachment selection field
- Since I already built out a way of listening for events in `editAtttributeField` in #336, it would make sense if this field could extend `editAttributeField` so that it could use the same system of wp js hooks.

That is separate from the question of whether the script for these fields is built through Browserify. I just noticed that it was an unnecessary step.

The only real interface the plugin offers, the `sui` and `sui.views` objects, are already accessible globally (when the field script is [enqueued](https://github.com/fusioneng/Shortcake/blob/master/inc/fields/class-field-attachment.php#L42), 'shortcode-ui' is registered as a dependency). I didn't understand the reasoning behind building an additional 10K of scripts into `build/field-attachment.js` to just to access those variables which were already accessible.

Plus, if we make it possible to extend the default editAttributeField without requiring devs to bundle chunks of scripts from this plugin, then its possible to extend Shortcake for capabilities that don't have to be built into the plugin itself. Otherwise, developers are limited to the fields provided.  

I could well be missing something. @jitendraharpalani58 do you have any feedback here?

jitendraharpalani58:
@goldenapples 

> I didn't understand the reasoning behind building an additional 10K of scripts into `build/field-attachment.js`

I definitely agree with you. The reason for creating a global variable `sui` was to access to plugin users to create their own custom fields. For example, we ( at Fusion ) too, had `post-select` in fusion-theme itself. Later it was moved to `shortcake`. 

Although, If these custom fields are making their way into `shortcake`, I would rather think of adding them under `/views`, and then, extend `editAttributeField` via browserfiy. This way they are appended in `shortcode-ui.js` and we don't have to make another http request for `field-attachment.js` at all.

goldenapples:
See issue #360.

goldenapples:
Following @jitendraharpalani58's comments, I moved the `editAttributeFieldAttachment` view into the main js bundle loaded by the plugin. The class itself is written as a standalone view which would work outside of a requirejs module just as well, if the initial `require` and the ending `module.exports` lines were removed. I'd like to keep to this pattern so that we can show plugin authors a clear example of how to extend the `editAttributeField` class.

<del>One thing I'm not sure about is how to store custom data needed by a view that should be accessible outside the plugin. I'm trying to expose the attachment cache globally, because plugins hooking into the attachment field's update event will need to be able to access that. I've just added a `.data` property to the `sui` global to hold this. But I don't really like this approach - it encourages developers to build in the globals rather than keep their data scoped to their own view classes. Is there a better convention for storing this kind of data?</del>

goldenapples:
> Is there a better convention for storing this kind of data?

I ended up just using class methods on the view to store the cache and get items from it or set items in it. Seems better to me.

js/src/field-attachment.js
```diff
@@ -1,152 +0,0 @@
-var sui = require('sui-utils/sui'),
-    editAttributeField = require( 'sui-views/edit-attribute-field' ),
-    $ = require('jquery');
-
-// Cache attachment IDs for quicker loading.
-var iDCache = {};
-
-sui.views.editAttributeFieldAttachment = editAttributeField.extend( {
```
danielbachhuber, line 8: What will the removal of this file break?

danielbachhuber:
I don't have any strong opinions on the architectural changes. I'm happy to be a rubber duck as you talk through them though.

It would be good to get @mattheu's input on this PR

js/src/field-attachment.js
goldenapples, line 8: Nothing. I'm not sure why Github is showing it as a removal, it's really just being moved to `js/src/views/edit-attribute-field-attachment.js` and built into the `shortcode-ui.js` bundle.

goldenapples:
OK, I'm starting to feel pretty good about this shape of this now.

I've added a couple of helper methods for dealing with attribute field views: `getValue()` and `setValue()` - just to avoid always having to look inside the model attached to the view to get or set those values.

I also added a finder function, `getField()` as a public class method of editAttributeField, to simplify the process of finding another view in a collection of attribute fields.

Finally, I added a class method to `editAttributeFieldAttachment` to get an attachment post from the cache: `getFromCache()`. This avoids the need to have another global exposed there.

An example of a listener using these helper methods is at https://github.com/fusioneng/image-shortcake/pull/44/files

@mattheu @danielbachhuber Do either of you want to review this or give me some feedback on the direction it's going? Now that the basic structure to extend is starting to fall into place, it should be simple to rework the other attribute field types so that they also inherit some of these helper methods and event triggers.

danielbachhuber:
:+1: from me

Issue #373, opened 2015-07-02
Can't get color picker to work


andrewdodson14:
I can't seem to get the color picker to save my selection and/or display on the front end. I also can't seem to find any real documentation on this new field. How should I go about using it?

danielbachhuber:
Can you share the code you've written in a Gist?

andrewdodson14:
Sorry for the slow response... here's my code:
https://gist.github.com/d47363f1073a07327845.git

andrewdodson14:
<?php

add_action( 'init', function() {

```
/**
 * Register your shortcode as you would normally.
 * This is a simple example for a pullquote with a citation.
 */
add_shortcode( 'custom_text', function( $attr ) {

    $attr = wp_parse_args( $attr, array(
        'text' => '',
        'element_tag' => 'p',
        'text_color' => 'black',
        'color_picker' => '',
        'font_weight' => '400-regular',
        'align' => 'center'
    ) );

    ob_start(); ?>

    <<?php echo $attr['element_tag']; ?> class="<?php echo 'custom-text text-' . $attr['align'] . ' ' . $attr['font_weight'] . ' '; if(! $attr['color_picker']) { echo $attr['text_color']; } ?>" <?php if($attr['color_picker']) { echo 'style="color: ' . $attr['color_picker'] . ';"'; } ?>>
        <?php echo $attr['text']; ?>
    </<?php echo $attr['element_tag']; ?>>

    <? $content = ob_get_contents();
    ob_end_clean();
    return $content;
} );

/**
 * Register a UI for the Shortcode.
 * Pass the shortcode tag (string)
 * and an array or args.
 */
shortcode_ui_register_for_shortcode(
    'custom_text',
    array(

        // Display label. String. Required.
        'label' => 'Custom Text',

        // Icon/image for shortcode. Optional. src or dashicons-$icon. Defaults to carrot.
        'listItemImage' => 'dashicons-editor-textcolor',

        // Available shortcode attributes and default values. Required. Array.
        // Attribute model expects 'attr', 'type' and 'label'
        // Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date.  
        'attrs' => array(
            array(
                'label' => 'Text',
                'attr'  => 'text',
                'type'  => 'text',
            ),
            array(
                'label'       => 'Element Tag',
                'attr'        => 'element_tag',
                'type'        => 'select',
                'options'     => array(
                    'p' => 'p',
                    'h1' => 'h1',
[...]

PR #377, opened 2015-07-07
Fix post select field

goldenapples:
The post select field was depending on the updateValue() method from `editAttachmentField`, whcih was changed in #358. This gives it its own listener function which gets the updated value from the select2 event.


goldenapples:
This fixes the "post select" attribute field, which was broken after changes in #358. A similar problem exists with the "color picker" field as noted in #373. I will open a separate pull request fixing that.

PR #379, opened 2015-07-07
Fix bugs in color picker field

goldenapples:
Fixes a couple problems with the color picker field:
- The model wasn't being updated after a change event in the
  colorpicker;
- The model's meta value wasn't being concatenated properly, causing
  HTML errors in the input element where any meta data was being output
  as `[object Object]`.


goldenapples:
See #373.

Still trying to track down some console errors that are showing up. Strangely enough, I'm trying to track down this bug with git bisect and I can't find a ref where the color-picker field actually worked...

davisshaver:
@goldenapples How's this going? Want me to take it over finish next week?

goldenapples:
@davisshaver That'd be great if you have the cycles to fix this up. I know how I wanted to tackle it but I'm deep in a different project right now...

goldenapples:
@danielbachhuber After the preview tab was removed in #411, the console error I was seeing no longer occurs. This should be ready for merge. (The build error looks like a problem with Travis, but I'm not able to restart the build. Can you restart the build to see if it passes?)

Issue #366, opened 2015-06-29
Use get_post_type() instead of get_current_screen()->post_type


danielbachhuber:
`get_post_type()` is more appropriate because it isn't fixed to specific admin contexts.

Previously #355 
Related #337

danielbachhuber:
:ship:

PR #381, opened 2015-07-09
Use `get_post_type()` instead of `get_current_screen()->post_type`

danielbachhuber:
The former works well in frontend context, where the latter only
works in the admin.

See #366

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -82,10 +82,10 @@ class Shortcode_UI {
 		wp_enqueue_media();
 
 		$shortcodes = array_values( $this->get_shortcodes() );
-		$screen = function_exists( 'get_current_screen' ) ? get_current_screen() : false;
-		if ( $screen && ! empty( $screen->post_type ) ) {
+		$current_post_type = get_post_type();
+		if ( $current_post_type ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@goldenapples Can you sanity check this?

goldenapples:
I guess I'm not sure what this is supposed to solve? It doesn't seem to fail spectacularly anywhere, but I don't know that it really gets the context you want either. On the front end, I assume this would be being used inside of comment boxes? Or maybe in some context where a plugin or theme tried to use a WP_Editor to give users a wysiwyg editor to fill out a form? In either of those cases, I don't think that a post type of "post" accurately describes the context to a point where it makes sense to choose which shortcodes to provide UI for.

danielbachhuber:
> I guess I'm not sure what this is supposed to solve?

Sorry, I should've been more clear.

>  On the front end, I assume this would be being used inside of comment boxes? Or maybe in some context where a plugin or theme tried to use a WP_Editor to give users a wysiwyg editor to fill out a form?

Correct, and correct.

> In either of those cases, I don't think that a post type of "post" accurately describes the context to a point where it makes sense to choose which shortcodes to provide UI for.

I think we can solve that problem when it's more of a problem.

PR #382, opened 2015-07-13
Fix undefined function JS error

mattheu:
I get a JS error when using the attachment field in the lates Shortcake. 

Looks like this didn't get updated following https://github.com/fusioneng/Shortcake/commit/d631032f387702c77d8df930a59f9799137b5be9


mattheu:
Review & Merge

danielbachhuber:
Previously #358

danielbachhuber:
And #377

Issue #385, opened 2015-07-14
Only register shortcode UI when editor is being created


jrfnl:
Lean loading - based on the current example code (dev.php), the UI would always be registered, no matter whether in front-end, back-end, non-editor admin page etc. as it hooks into `init`

I'd like to suggest creating a hook people should use to register the shortcode UI on and having that hook only run when an editor is being created.

So then it would become something along the lines of:

``` php
add_action( 'register_shortcode_ui', 'my_plugin_shortcode_ui_registration_function' );

function my_plugin_shortcode_ui_registration_function() {
    shortcode_ui_register_for_shortcode(
        'shortcode_1',
        array(...)
    );
    shortcode_ui_register_for_shortcode(
        'shortcode_2',
        array(...)
    );
}
```

This would also remove the need for the `function_exists( 'shortcode_ui_register_for_shortcode' )` checks as the hook will only be run from Shortcake and therefore the function will be available.

danielbachhuber:
> I'd like to suggest creating a hook people should use to register the shortcode UI on and having that hook only run when an editor is being created.

I dig it #388

cggit:
admin_init seems to work but Im having trouble getting the current page ID so early. If you go the custom hook route can you pass it some variables so we can have some context.

danielbachhuber:
`register_shortcode_ui` will be the trick.

PR #388, opened 2015-07-15
Offer a `register_shortcode_ui` hook for plugins to register on

danielbachhuber:
See #385

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -27,9 +27,14 @@ class Shortcode_UI {
 	}
 
 	private function setup_actions() {
+		add_action( 'init',                  array( $this, 'action_init' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@goldenapples how do you feel about this?

goldenapples:
This doesn't seem like it addresses the issue @jrfnl raised, as the shortcode ui is still being registered on `init`; which includes all front- and back-end views.

What about firing the `register_shortcode_ui` action inside of the `action_admin_enqueue_scripts()` function? That seems to be the only context you'd need shortcode UI registered for, and I believe that's early enough to hook in and register scripts and styles...?

danielbachhuber:
Shouldn't you be able to use shortcode UI on the frontend?

danielbachhuber:
I suppose we could somehow hook into when `wp_editor()` is called.

goldenapples:
... which would be `wp_enqueue_editor`, no? Sorry, I suggested the wrong function. `action_enqueue_wp_editor` would be a more appropriate place to fire the hook.

jrfnl:
Nice ;-)
Maybe update the dev.php example code to mirror the new implementation ?

danielbachhuber:
Closing in favor of #429

PR #391, opened 2015-07-20
inline editor experiment

mattheu:
Experimental... do not merge.

I've been looking into the plausability of inline editing.
1. Requires providing a completely custom view for the whole form, not just single fields.
   - This would probably be nice to have anyway.
2. Requires a core patch to allow you to set the content of an MCE view as a jQuery object. This allows all the events defined in the view to work. [Get it here](https://dl.dropboxusercontent.com/u/21013057/test-jquery-objects.diff)
3. NOT WORKING - saving. Not sure how to do this... using send_to_editor will refresh the MCE view and disrupt your editing experience.  

Thoughts on 3.
You can probably work around and handle updating the content on save. I did get something working but it rendered the shortcode twice. This could also cause issues with multiple editors...


khromov:
Very nice initiative. Inline editing would make this an amazing experience for content editors. :+1:

Issue #392, opened 2015-07-20
Start using the regex that WordPress uses in PHP for shortcode parsing


nikolov-tmw:
Right now a very simple regular expression is used for parsing shortcodes, compared to what WordPress natively uses in `get_shortcode_regex()`. This breaks things like shortcodes within the `inner_content`(which is more or less a common use case).

I've managed to import the regex over to JS and I'm going to create a pull request here with the adjustments I've done.

danielbachhuber:
:+1:

khromov:
@nikolov-tmw Was there any issues with this? Don't see a PR for it.

nikolov-tmw:
It's right here - https://github.com/fusioneng/Shortcake/pull/393 .
I'm supposed to write some Unit tests, but never got around to it. I'll try to allocate some time next week to write the tests and update(if necessary) the regex, since there could have been changes to it in WordPress.

khromov:
@nikolov-tmw Thanks Nikola, don't know how I missed that! 

The PR looks solid, looking forward to it getting merged!

danielbachhuber:
I'll see what I can do about bringing #393 over the finish line with test coverage when I have the time.

khromov:
@nikolov-tmw  I've been looking some more at this, I don't think there's a need to bring a whole new regexp into this, as there is already a JS version of the shortcode regexp, available here:
https://github.com/WordPress/WordPress/blob/master/wp-includes/js/shortcode.js#L105

I imported it into a separate branch and the tests pass fine. We could even use the `wp.shortcode.regexp( tag )` constant, although the tests don't pass when using it (since I guess the tests don't include WP core JS functions)

danielbachhuber:
:100: for using the JS version of shortcode regex.

@nikolov-tmw shall I leave for you to open a new PR with reasonable test coverage?

danielbachhuber:
:ship:

PR #393, opened 2015-07-21
Addresses #392 - importing the WordPress shortcode regex

nikolov-tmw:
Fixes the issue where the inner content would be partially or fully truncated when editing a shortcode that contains another shortcode.

Without the fix:

Shortcode preview in rich text

Last part "amet." is missing from the Content.
"amet." missing

With the fix:

Full content is preserved

Below is a minimal plugin that allows you to replicate the issue and the fix. We have one shortcode( **simple** ) that just runs the content through `do_shortcode()` and another one that just puts the content in `<b></b>` tags( **b** ). 

```
<?php
/**
 * Plugin Name: Shortcodes Test
 */

function sct_init() {
    add_shortcode( 'simple', 'sct_simple_shortcode' );

    add_shortcode( 'b', 'sct_b_shortcode' );

            'simple',
            array(
                'label'         => 'Plain Content',
                'listItemImage' => 'dashicons-welcome-write-blog',
                'inner_content' => array(
                    'label' => 'Content',
                ),
                'attrs'         => array(),
            )
        );
    }

}
add_action( 'init', 'sct_init', 10 );

function sct_simple_shortcode( $atts, $content = null ) {
    extract( shortcode_atts(array(), $atts, 'simple' /* Shortcode name for attribute filtering */ ) );

    return do_shortcode( $content );
}

function sct_b_shortcode( $atts, $content = null ) {
    extract( shortcode_atts(array(), $atts, 'b' /* Shortcode name for attribute filtering */ ) );

    return '<b>' . $content . '</b>';
}
```

### DIFF
--- a/js/build/field-color.js
+++ b/js/build/field-color.js
@@ -1,6 +1,6 @@
 (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
 (function (global){
-var Backbone = (typeof window !== "undefined" ? window.Backbone : typeof global !== "undefined" ? global.Backbone : null);
+var Backbone = (typeof window !== "undefined" ? window['Backbone'] : typeof global !== "undefined" ? global['Backbone'] : null);
 var ShortcodeAttribute = require('./../models/shortcode-attribute.js');
 
 /**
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@nikolov-tmw This would really benefit from test coverage. Would you be willing to add it?

nikolov-tmw:
Not sure how to do that :eyes: I haven't really written any JS tests before.
I'm also not sure what we should test for - make sure the content remains intact with and without shortcodes(with shortcodes with/without attributes). Anything else?

danielbachhuber:
> Not sure how to do that :eyes: I haven't really written any JS tests before.

@noppanit can correct me if I'm wrong, but I think https://github.com/fusioneng/Shortcake/blob/master/js-tests/src/shortcodeModelSpec.js is a good place for it. He's written up some documentation here: https://github.com/fusioneng/tech-docs/blob/d47d1852d364a6dcae083e597658d289a22f17d8/wordpress-development/javascript-tests.md

> make sure the content remains intact with and without shortcodes(with shortcodes with/without attributes)

Yep, that'd be good. Generally, testing "both sides" of a bug is good practice.

nikolov-tmw:
@danielbachhuber @noppanit - I see what the tests look like and it seems fairly easy to add new tests. However I wanted to ask if I should create a different instance of `Shortcode()` for every different content test.

Because a good place for the new tests would be inside of the 'Inner content set correctly..' section.

Or should I just add for instance

`shortcode.set( 'inner_content', 'test content [test]with shortcodes[/test]' )`

And then do

`expect( shortcode.get( 'inner_content' ).get('value')...`

danielbachhuber:
> However I wanted to ask if I should create a different instance of Shortcode() for every different content test.

You can just reuse the existing model.

khromov:
@nikolov-tmw There seems to be an issue if you nest a shortcode that is also registered in Shortcake. Shortcodes that are not registered in Shortcake work fine when you nest them, but if they are registrered in Shortcake everything breaks.

danielbachhuber:
This was landed in #487

PR #400, opened 2015-07-26
Fixes an issue where editing a shortcode with newlines in attributes would only populate the first line of the attribute's textarea.

joeyblake:
Changes the regex for individual shortcode atts to allow for newlines in the attributes.


davisshaver:
@joeyblake, I'm cleaning up the diff for possible merge in https://github.com/fusioneng/Shortcake/pull/418. I think you had a different version of a grunt plugin, causing the JS diff.

davisshaver:
Closing in favor of #418.

PR #405, opened 2015-07-30
Add range input type

davisshaver:


js-tests/build/specs.js
```diff
@@ -58,7 +58,7 @@ var Shortcode = require('../../js/src/models/shortcode');
 var InnerContent = require('../../js/src/models/inner-content');
 var ShortcodeAttribute = require('../../js/src/models/shortcode-attribute');
 var ShortcodeAttributes = require('../../js/src/collections/shortcode-attributes');
-var $ = (typeof window !== "undefined" ? window.jQuery : typeof global !== "undefined" ? global.jQuery : null);
+var $ = (typeof window !== "undefined" ? window['jQuery'] : typeof global !== "undefined" ? global['jQuery'] : null);
```
davisshaver: @noppanit any idea what might cause this diff? I believe my npm modules are up to date, just ran `npm update`.

noppanit: I suspect it's the version of browserify. What version is on your system right now? The one in this repo is `3.4`

davisshaver: @noppanit I don't think so, just tried installing `grunt-browserify` at 3.4 and no diff generated.

Would you mind running locally and seeing if you generate a diff from this branch? If so, maybe you can push it? The change seems innocuous, I just don't understand why it's happening on this branch.

goldenapples:
I had a similar problem a while ago, caused by having a more recent version of browserify-shim. I fixed in https://github.com/fusioneng/fusion-theme/commit/d83c08cf05b87967859b8a9e443b513b3a4ee90d by pinning the version to the last one which doesn't cause this change. can you try that?

davisshaver:
@goldenapples Much cleaner, thank you. Mind giving me some code review, too?

inc/templates/edit-form.tpl.php
```diff
@@ -135,3 +135,24 @@
...
+			<output class="range" for="{{ data.id }}" id="{{ data.id }}_indicator">{{ data.value}}</output>
+		</div>
+		<# if ( typeof data.description == 'string' ) { #>
+			<p class="description">{{ data.description }}</p>
+		<# } #>
+	</div>
+</script>
+
+<script>
+	function updateRangeIndicator( event ) {
```
goldenapples: Why does this function have to be defined in the templates file? It seems like it makes more sense as an event handler on the `editAttributeField` view.

davisshaver: Why? Because I don't really grok backbone yet :). I'll attempt a refactor though...

davisshaver:
@goldenapples Moved this into the existing `inputChanged` function, which already has some type specific functionality, so it seems to be the right place.

js/src/views/edit-attribute-field.js
```diff
@@ -71,6 +72,10 @@ var editAttributeField = Backbone.View.extend( {
 			this.setValue( $el.filter(':checked').first().val() );
 		} else if ( 'checkbox' === this.model.attributes.type ) {
 			this.setValue( $el.is( ':checked' ) );
+		}  else if ( 'range' === this.model.attributes.type ) {
+			var rangeId =  '#' + e.target.id + '_indicator';
+			var rangeValue = e.target.value;
+			document.querySelector( rangeId ).value = rangeValue;
```
goldenapples, line 78: You still need to call `this.setValue()`, otherwise the underlying model won't be updated and won't save properly.

goldenapples, line 78: Also `e.target` should already be defined as `$el` here. Might be easier to use that instead?

davisshaver, line 78: @goldenapples I think the string `id` is easier to use than the object because I'm constructing the selector. Updating with `this.setValue()` now.

davisshaver:
@goldenapples merge?

goldenapples:
> Moved this into the existing inputChanged function, which already has some type specific functionality, so it seems to be the right place.

Yes, that approach looks much better to me. I originally thought it would be useful to give this field its own view model that extends `editAttributeField`, but since its such a tiny amount of type-specific code, it fits pretty well in there. And it seems useful to have one view model that covers all the basic HTML input types.

davisshaver:
@goldenapples I tried that first (as you can see in commits) but then I was cutting off the other event, so agreed this seems to be the "right" place for the code.

goldenapples:
OK. One final critique/question: I liked the feel of the slider much better when you were updating the indicator on `input`, rather than on `change`. What do you think about either:
- binding another event on `input input[type="range']` which updates the indicator; then you could remove all the type-specific code from `inputChanged` for this field type and let it just `setValue` like any other field types do.
- changing the event bound to `inputChanged` for this field type to `input`. This would mean the model was updated, and an action triggered, every step of the slider, which could be used to enable nice UI, like showing image dimensions at different percentages or updating a preview size.

davisshaver:
The latter seems preferable to me. Different inputs throw different events that are better/worse for certain uses. I think changing it just for the range is ok... Updated.

goldenapples:
Nice, looks good to me now. Merging.

PR #406, opened 2015-07-31
Update range output on `input` event

goldenapples:
Enables smoother indicator updating, rather than waiting until selection
stops.

@davisshaver this is something like what I had in mind...


davisshaver:
Closing as this was handled in original PR https://github.com/fusioneng/Shortcake/pull/405/files#diff-7b22e98c6a919f8d9d9e44eea1ed4279R234

PR #409, opened 2015-08-03
Indicate attachment metadata

davisshaver:
See https://github.com/fusioneng/Shortcake/issues/351

Bring Shortcake experience
screen shot 2015-08-03 at 4 11 22 pm

Closer to Core
screen shot 2015-08-03 at 4 07 55 pm


davisshaver:
@goldenapples Sorry to pile on :/. This PR injects a `thumbnail-details` element into a new `thumbnail-details-container` that has some basic attachment metadata.

js/build/shortcode-ui.js
```diff
@@ -721,6 +723,40 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
...
+			class: 'dimensions',
+			html: attachment.height + ' &times; ' + attachment.width ,
+		}).appendTo( $thumbnailDetails );
+
+		jQuery( '<div/>', {
+			class: 'edit-link',
+			html: '<a href="' + attachment.editLink + '">Edit Image</a>',
+		}).appendTo( $thumbnailDetails );
+
+		this.$metaContainer.append( $thumbnailDetails );
```
goldenapples: I'm not a big fan of all this jQuery DOM-building going on here in the view. Seems like this is what templates are for. 

Is there a way to rework this field so that it makes better use of templates, both here and in the image preview?

danielbachhuber: > I'm not a big fan of all this jQuery DOM-building going on here in the view.

In addition, VIP won't permit use of `html()` to set arbitrary HTML. For some fun sleuthing as to why, check out the [WordPress 4.2.4 release notes](https://wordpress.org/news/2015/08/wordpress-4-2-4-security-and-maintenance-release/), in particular how [r33541](https://core.trac.wordpress.org/changeset/33541/branches/4.2) could be exploited.

goldenapples:
The output looks good, but my gut feeling is that it'll be a mess to maintain. If this could be reworked either to include these fields in the `tmpl-fusion-shortcake-field-attachment` template or in a sub-template which you call in the view, that would seem much cleaner to me. I think the functions rendering the attachment preview are already too messy here, I wouldn't want to add any more DOM-building...

davisshaver:
@goldenapples I refactored to bootstrap the existing template, and then just hide/populate text as needed. I think if we want to go to an actual Backbone template maybe we should refactor how the image preview gets constructed. What do you think about this? We could also add as a sub-template, but I'd probably need some more pointers on that.

inc/fields/class-field-attachment.php
```diff
@@ -64,6 +64,14 @@ public function action_shortcode_ui_loaded_editor() {
 						<div class="attachment-preview-loading"><ins></ins></div>
 					</div>
 				</div>
+				<div class="thumbnail-details-container">
+					<strong>Thumbnail Details</strong>
```
danielbachhuber: This string should be localizable

js/build/shortcode-ui.js
```diff
@@ -722,6 +723,13 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
 		this.$container.append( $thumbnail );
 		this.$container.toggleClass( 'has-attachment', true );
 
+		this.$thumbnailDetailsContainer.find( '.filename' ).text( attachment.filename );
+		this.$thumbnailDetailsContainer.find( '.date-formatted' ).text( attachment.dateFormatted );
+		this.$thumbnailDetailsContainer.find( '.size' ).text( attachment.filesizeHumanReadable );
+		this.$thumbnailDetailsContainer.find( '.dimensions' ).text( attachment.height + ' × ' + attachment.width );
+		this.$thumbnailDetailsContainer.find( '.edit-link a' ).attr( "href", attachment.editLink );
+		this.$thumbnailDetailsContainer.toggleClass( 'has-attachment', true );
```
danielbachhuber, line 731: Good :+1:

davisshaver:
@danielbachhuber Updated for translation

goldenapples:
> What do you think about this? We could also add as a sub-template, but I'd probably need some more pointers on that.

This reads much better to me. Lets tackle reworking templates in a separate issue. I'd be into merging this as is. @danielbachhuber see anything else?

danielbachhuber:
Looks good to me :+1:

PR #410, opened 2015-08-04
Better UX for changing an attachment

davisshaver:
See https://github.com/fusioneng/Shortcake/issues/350


davisshaver:
@goldenapples Seemed almost too easy, what would you add?

goldenapples:
:100: 

It seems like a hover state on the thumbnail div might be nice too, to aid in discoverability. I'd try an overlay or something?

davisshaver:
Thanks @goldenapples, good idea. I added a light overlay to the image on hover to aid in discoverability. Right now it's `rgba( 0,0,0,.1)`, I could see us bumping that up maybe but I wasn't sure.

aug 05 2015 07 41

goldenapples:
Nice! I like this look, and the experience is really smooth...

Issue #66, opened 2014-12-02
Capture keyboard bindings when modal is open


danielbachhuber:
When I have the modal open to edit a shortcode:

![image](https://cloud.githubusercontent.com/assets/36432/5265543/fae4feb0-79f7-11e4-832c-36ceee12f8b5.png)

We should capture a couple of keyboard bindings:
- [x] ESC should close the modal, without updating.
- [ ] Enter / return should close the modal, with updating.

danielbachhuber:
@mattheu and I spent some time on this today. Core actually handles the ESC case for us — when the modal has focus. The problem is that the modal doesn't have focus when initially loaded, because the TinyMCE plugin steals it back: https://core.trac.wordpress.org/browser/trunk/src/wp-includes/js/tinymce/plugins/wpview/plugin.js#L331

Because document focus is within the iframe, the parent can't steal focus back. Let's punt on this until we get some resolution from core about it.

danielbachhuber:
Opened a core ticket https://core.trac.wordpress.org/ticket/30817

danielbachhuber:
Just need to handle the `return` case.

PR #412, opened 2015-08-04
First pass at capturing return key events

davisshaver:
See https://github.com/fusioneng/Shortcake/issues/66


davisshaver:
Having trouble with element focus... the `keyup` isn't being triggered unless an element within the form has focus.

davisshaver:
@goldenapples @noppanit Could I get maybe some pair-programming or eyes on this PR?

I want to capture `keyup` but right now the event is only firing when an input is selected. It looks like maybe this is a known issue with Backbone: http://stackoverflow.com/posts/20720819/revisions

Any ideas?

goldenapples:
I'd assume that unless the focus is inside the view's element, that `keyup` won't bubble through that element, and therefor you won't be able to catch it on the view. You probably would need to attach a `keyup` listener on `document` or `window` which, if necessary, triggers a custom event on `EditShortcodeForm`.

goldenapples:
Adding a listener like this to the `initialize` function will capture all <kbd>return</kbd> keypresses.

```
jQuery('body').on('keyup', function(e) { t.possiblySelect(e); });
```

Course, you'd really need to namespace it, and unbind it on closing the modal, and some other cleanups, but I think this might be the only way to listen for this event. It's not necessarily going to bubble through the view otherwise...

davisshaver:
Thanks @goldenapples, I was worried that'd be the case. I'll try this with jQuery calls in `initialize` and `teardown`

davisshaver:
I've got this started, but after digging into it, I don't think we should pursue for merge at this point.
- This UX (insert on <kbd>return</kbd>) is not in parity with the Core media modal experience. 
- One reason this might be is that <kbd>return</kbd> could be a legitimate input for attribute fields within a shortcode. Richtext fields, for instance.
- An option would be to check which element has focus, and only running the function. This branch `console.log`'s the element with focus on `keyUp`. If you check it out, you'll notice that focus doesn't seem predictable as the user navigates around the modal context. And we haven't even got into devices yet.
- The UX is also undesirable as Shortcake has no functions for input/shortcode validation at this point. If we wanted to quick insert on <kbd>return</kbd>, after playing around with this idea, I think we'd need validation abstraction first.

@danielbachhuber @goldenapples Thoughts?

danielbachhuber:
I'm open to :football: — I appreciate all of the discovery you've done into this.

Issue #359, opened 2015-06-20
Pass false shortcode value for unchecked checkbox


niallkennedy:
A shortcode includes a boolean attribute, do_the_thing, with a default state of true. A falsey attribute value overrides the default.

The shortcode author exposes the attribute to Shortcake through a `checkbox` field type, passing a `value` of `true` in string form, resulting in a checked property output in `tmpl-shortcode-ui-field-checkbox`. The default value of the shortcode is communicated.

The viewer unchecks the checkbox, overriding the default.

If Shortcake built an attribute value of `false` for the unchecked checkbox state the author's shortcode handler could respond to falsey values as well as truthy values for the generated shortcode.

The existing template behavior requires a string of `true` to set the truthy state. A shortcode could be similarly expected to respond to a string of `false` generated by Shortcake to represent the falsey state.
### Example shortcode result

```
[handle do_the_thing="false"]
```
### Example registered shortcode fragment

``` php
'attrs' => array(
  array(
    'attr'  => 'do_the_thing',
    'label' => 'Do the thing?',
    'type'  => 'checkbox',
    'value' => 'true',
  ),
)
```

danielbachhuber:
Thanks again for the report, @niallkennedy. We'll get this fixed for the next release.

danielbachhuber:
Holding on this change for v0.5.0, as I don't yet feel 100% confident about this change.

khromov:
This was a recurring issue with Visual Composer a while back.

There are three states:
- Checked (param="true")
- Unchecked (param="false")
- Undefined (Not yet checked or unchecked. No param)

Shortcake needs to support all three of these in order for the checkbox to work in a meaningful way.

astratagem:
Yes this makes checkboxes useless in some cases where we need to check for a `false` value. It's easy to work around with a select field and defining `true` and `false` keys but checkboxes are better suited for boolean values.

PR #413, opened 2015-08-05
Pass explicit `false` value for falsey checkboxes

davisshaver:
Checkbox values will be passed to the shortcode for both `true` and `false` states.

See https://github.com/fusioneng/Shortcake/issues/359

### DIFF
--- a/js-tests/src/shortcodeModelSpec.js
+++ b/js-tests/src/shortcodeModelSpec.js
@@ -18,6 +18,11 @@ describe( "Shortcode Model", function() {
 				type:        'text',
 				value:       'test value',
 				placeholder: 'test placeholder',
+			}, {
+				attr:        'checkbox',
+				label:       'Attribute',
+				type:        'checkbox',
+				value:       'true',
 			}
 		],
 		inner_content: {
@@ -38,8 +43,9 @@ describe( "Shortcode Model", function() {
 
 	it( 'Attribute data set correctly..', function() {
 		expect( shortcode.get( 'attrs' ) instanceof ShortcodeAttributes ).toEqual( true );
-		expect( shortcode.get( 'attrs' ).length ).toEqual( 1 );
+		expect( shortcode.get( 'attrs' ).length ).toEqual( 2 );
 		expect( shortcode.get( 'attrs' ).first().get('type') ).toEqual( data.attrs[0].type );
+		expect( shortcode.get( 'attrs' ).last().get('type') ).toEqual( data.attrs[1].type );
 	});
 
 	it( 'Inner content set correctly..', function() {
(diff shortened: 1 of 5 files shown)


danielbachhuber:
Hmm. This seems like one of those things it would be really helpful to have a test case for.

How should I review? Do we need to accommodate `type=radio` as well?

davisshaver:
@danielbachhuber No, I don't think we need to accommodate the radio button.

What we've changed here is how a truthy default works with the checkbox. 

Before this fix, assuming `prop` is a checkbox attribute with a truthy default, if the user unchecked `prop` then there'd be a loss of information as `[shortcode]` would be passed back to TinyMCE. Then when the checkbox field was rendered again, the default would again override the false value for the field.

 If the default of a radio is some 'value', then we expect the shortcode to look like `[shortcode prop="value"]` when it's passed back to the editor. This value would be reflected in the input and all would be good. If no default value is provided then the functional default is no selection, which would just mean that the attribute doesn't get passed to the shortcode, but this is fine as the input will render with no selection anyway.

I've added a test based on @niallkennedy's use case.

davisshaver:
These gifs aren't good quality but they demonstrate how this PR improves the use case.

Before
ugliergif

After
uglygif

js-tests/src/shortcodeModelSpec.js
```diff
@@ -58,16 +64,19 @@ describe( "Shortcode Model", function() {
...
+		expect( _shortcode.formatShortcode() ).toEqual( '[test_shortcode attr="test value" checkbox="true"]' );
 
 		// Test without attributes
 		_shortcode.get( 'attrs' ).first().unset( 'value' );
-		expect( _shortcode.formatShortcode() ).toEqual( '[test_shortcode]' );
+		expect( _shortcode.formatShortcode() ).toEqual( '[test_shortcode checkbox="true"]' );
 
+		// Test with falsey checkbox
+		_shortcode.get( 'attrs' ).last().set( 'value', 'false' );
+		expect( _shortcode.formatShortcode() ).toEqual( '[test_shortcode checkbox="false"]' );
```
danielbachhuber, line 79: I'd expect this to be `checkbox=""`...

danielbachhuber:
@davisshaver I think we should chat about this some during the Shortcake chat today. I'm not 100% sure this is the behavior we want.

js-tests/src/shortcodeModelSpec.js
davisshaver, line 79: @danielbachhuber That would evaluate as null, which would trigger the default to override (Which is the behavior we are trying to prevent).

davisshaver:
@danielbachhuber Sure, fine with me. 

Maybe we could ask @niallkennedy what he expects, too.

niallkennedy:
I look for falsey values (0, 'false', etc) for a shortcode override. I ignore empty string values for a boolean attribute.

The checked behavior passes a 'true' string to be interpreted by the shortcode handler. If you support Shortcake you also need to accept a value of 'true' for a boolean shortcode attribute. A string of 'false' represents the opposite state in my mind, and provides consistency between the required shortcode attribute values a shortcode handler should accept to be compatible with Shortcake.

davisshaver:
Thanks @niallkennedy!

davisshaver:
@danielbachhuber based on https://github.com/fusioneng/Shortcake/pull/413#issuecomment-129570778, what do you think about merging? Should we get a second opinion from @mattheu or @goldenapples?

davisshaver:
@danielbachhuber Assigning this to you for a second opinion – still hesitant about behavior change?

danielbachhuber:
> still hesitant about behavior change?

Yes. For now, let's just bump it from the v0.5.0 milestone.

astratagem:
It would be great to see this merged at some point. And if not, at least an explanation of why this might be unwanted behavior and possible workarounds? Currently I'm just avoiding checkbox fields because of this and using select fields instead.

goldenapples:
I like this concept, but it seems that we would need some way of enabling this as an opt-in feature on an attribute. I know that a number of shortcodes I've built would break if this were merged as-is, as there are plenty of places where I just check for a boolean attribute to be present in the callback, and `attribute="false"` would evaluate to true in the callback.

It might be good to follow the pattern we established with the "encoded" parameter, where a flag set on an attribute could be used in a filter on "shortcode_atts_{$shortcode_tag}" [like this one is](https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L123) to parse boolean attributes and either unset the attribute or set it to a boolean false rather than a string.

Issue #390, opened 2015-07-17
updateAttachment doesn't error out on missing attachments


paulkarpenko:
If I attach an image to a shortcode, then delete that image from the media library, the shortcode edit screen does not give me the ability to clear out the attachment or replace it. The attachment preview block simply sits in the loading state forever. 

Moving $container.removeClass( 'loading' ); out of the .done() promise and into an .always() promise will fix the issue.

danielbachhuber:
Thanks for the report, @nvarchar. We'll get this fixed.

danielbachhuber:
:hammer:

PR #417, opened 2015-08-10
When attachment is missing, allow a new attachment to be selected

davisshaver:
See https://github.com/fusioneng/Shortcake/issues/390

New behavior:
new behavior


davisshaver:
Good for review/merge, pretty simple change with the jQuery promise

PR #420, opened 2015-08-13
Set frame selection to current attachment when applicable

danielbachhuber:
See #350


js/build/shortcode-ui.js
```diff
@@ -740,6 +740,24 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
 	_openMediaFrame: function(e) {
 		e.preventDefault();
 		this.frame.open();
+		if ( this.model.get( 'value' ) ) {
+			var query_args = {
+				'post__in': [ this.model.get( 'value' ) ],
+				'orderby': 'post__in',
+				'perPage': 1
+			};
+			var attachments = wp.media.query( query_args );
```
danielbachhuber: @goldenapples This is producing 0 attachments:

image

Any idea why?

goldenapples: Not sure. I spent a while digging through the parameters available to `wp.media.query`, but I can't seem to make any of them actually work:

screen shot 2015-08-13 at 11 04 54 am

I'll come back to this later, but I'm not getting it right now...

danielbachhuber: Dang. I'll take another look at it.

danielbachhuber: https://github.com/alleyinteractive/wordpress-fieldmanager/pull/379 might work instead

Issue #401, opened 2015-07-28
JS error if $content but no 'inner_content' arg


cliffordp:
If the shortcode's `args` array does not have `inner_content` argument but the Editor has content between opening and closing shortcode tags, clicking on Shortcake's "Edit" button (pencil icon) causes a JavaScript console error for https://github.com/fusioneng/Shortcake/blob/fed5de982e437835c6b204d7dc495a883e091b40/js/build/shortcode-ui.js#L455 and clicking Shortcake's Edit button does not open the shortcode UI.

Error message: `Uncaught TypeError: Cannot read property 'set' of undefined`
<img width="927" alt="screenshot 2015-07-28 11 35 03" src="https://cloud.githubusercontent.com/assets/1812179/8937224/b83c7080-351c-11e5-9e90-99663384cbf1.png">

Example with my custom `mdl-badge` shortcode:
- `[mdl-badge badge_text="Inbox" data="7"]` is ok (self-enclosed shortcode) **<-- how this particular shortcode is designed**
- `[mdl-badge badge_text="Inbox" data="7"][/mdl-badge]` is ok (`$content` == '')
- `[mdl-badge badge_text="Inbox" data="7"]hello[/mdl-badge]` causes JS error (`$content` == 'hello')

The shortcode's rendering still works, just not the Edit button.

So the issue is if someone manually creates/edits shortcodes in Text editor, they could shoot themselves in the foot and not know how to fix, since it's not obvious the issue is because there's `$content` but no `inner_content` argument

danielbachhuber:
Confirmed. I suppose we have a similar problem where attributes will be dropped if UI isn't registered for them.

PR #424, opened 2015-08-13
When no `inner_content` is registered, back up to hidden model attribute

danielbachhuber:
One approach for #401
Merges into #422


danielbachhuber:
@goldenapples How do you feel about this? I think I prefer the approach over #423

goldenapples:
I agree. If the shortcode doesn't register an inner_content field, it doesn't make sense to show the textarea on the edit screen for it. This approach preserves any content which may have been inserted into the inner_content field, without exposing it in the UI, which seems like the desired behavior.

PR #427, opened 2015-08-17
Add basic multiple support for attachment field

oritiger:
``` php
array(
    'label' => 'Attachment',
    'attr'  => 'attachment',
    'type'  => 'attachment',
    'libraryType' => array( 'image' ),
    'addButton'   => 'Select Image',
    'frameTitle'  => 'Select Image',
    'multiple' => true,
),
```

multiple


StaggerLeee:
I do understand you remove something from plugin core.
But cannot understand you seem to think we can read minds. Following this fact I make my own documentation, as official is almost non existent.

Back to this specific issue.
- What problems prevented you to implement "'multiple' => true" for attachments?
- Do you plan to bring it back ? If yes some estimate in how many months.
- If not say it just we know it.

Issue #428, opened 2015-08-23
How can I get the current page id to be used in shortcode_ui_register_for_shortcode


cggit:
I would like to use the current page ID in shortcode_ui_register_for_shortcode but it seems to hook too early and when I try to use the URL shortcake JS crumbles.

Any suggestions on how I can get/use the current page ID consistently while registering my UI.

danielbachhuber:
On master, and in v0.5.0 soon, you'll be able to use the `register_shortcode_ui` action to register shortcode UI. This should be late enough for you to get what you need.

However, to make sure I'm providing good advice, could you describe your use case in greater detail?

cggit:
I was populating a shortcode UI select box with post meta and needed the postID. I managed to get it by parsing the QUERY_STRING on 'after_setup_theme' and storing it to be available on 'init'. Not pretty but it works. Sounds like register_shortcode_ui will make it much easier. Awesome plugin.

PR #429, opened 2015-08-24
Conditionally call register_shortcode_ui action, only when an editor is actively enqueued

goldenapples:
#385

This calls the `register_shortcode_ui` hook only when a WP_Editor is actually enqueued on the current screen.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -65,6 +65,7 @@ class Shortcode_UI {
 	 * Setup plugin actions.
 	 */
 	private function setup_actions() {
+		add_action( 'init',                  array( $this, 'action_init' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );
 		add_action( 'wp_enqueue_editor',     array( $this, 'action_wp_enqueue_editor' ) );
 		add_action( 'wp_ajax_do_shortcode',  array( $this, 'handle_ajax_do_shortcode' ) );
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -65,7 +65,6 @@ class Shortcode_UI {
 	 * Setup plugin actions.
 	 */
 	private function setup_actions() {
-		add_action( 'init',                  array( $this, 'action_init' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );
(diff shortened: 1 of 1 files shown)


goldenapples:
@danielbachhuber Do you have thoughts on this, as an alternate approach to #388? This approach seem to better solve the problems discussed in #385 and #428, in that it fires later in the page load, and only when necessary to actively set up shortcode UI.

inc/class-shortcode-ui.php
```diff
@@ -68,6 +68,31 @@ private function setup_actions() {
...
+		if ( ! did_action( 'register_shortcode_ui' ) ) {
+
+			/**
+			 * Register shortcode UI for shortcodes.
+			 *
+			 * Can be used to register shortcode UI only when an editor is being enqueued.
+			 *
+			 * @param array $settings Settings array for the ective WP_Editor.
+			 */
+			do_action( 'register_shortcode_ui', $settings, $editor_id );
```
danielbachhuber, line 92: Is there a reason you might want to register different shortcode UI for different editors?

We could make the action more specific to accommodate this (`register_shortcode_ui_for_editor`), and make it clear the action could be fired multiple times on one page.

danielbachhuber:
> Do you have thoughts on this, as an alternate approach to #388?

Aside from my comment, this seems much cleaner :+1:

inc/class-shortcode-ui.php
goldenapples, line 92: > Is there a reason you might want to register different shortcode UI for different editors?

The only reason I could think of has to do with the fact that you may want to enable certain UIs only when the editor settings indicate that media buttons are enabled for the current editor, or that quicktags or tinymce is enabled.

True, these settings aren't useful here, because this action is only fired once, for the first editor on a page.

> We could make the action more specific to accommodate this (register_shortcode_ui_for_editor), and make it clear the action could be fired multiple times on one page.

Yeah, I thought about this, but Shortcake currently stores its data globally, in `Shortcode_UI::get_instance()->$shortcodes`. Does it make sense to store a set of shortcodes for each editor on a page? I could see the value in it in some cases, but in 99% of use cases that will just be duplicated data.

danielbachhuber, line 92: > Yeah, I thought about this, but Shortcake currently stores its data globally, in Shortcode_UI::get_instance()->$shortcodes. Does it make sense to store a set of shortcodes for each editor on a page? I could see the value in it in some cases, but in 99% of use cases that will just be duplicated data.

Oh, right. I suppose we'd need to refactor a bit further.

Issue #295, opened 2015-04-22
Enforce coding standards for JS


mattheu:
Following on from - https://github.com/fusioneng/Shortcake/pull/290#issuecomment-95245184 - we should have a grunt task that scans JS files.

noppanit:
@mattheu you thinking JSLint?

danielbachhuber:
:ship:

PR #430, opened 2015-08-24
Clean up JS according to WP coding standards

danielbachhuber:
See #295

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -81,8 +81,8 @@ describe( "Shortcode Model", function() {
 		},
 	};
 
-	var defaultShortcode = new Shortcode();
-	var shortcode = new Shortcode( data );
+	defaultShortcode = new Shortcode();
+	shortcode = new Shortcode( data );
 
 	it( 'Defaults set correctly.', function() {
 		expect( defaultShortcode.get( 'label' ) ).toEqual( '' );
@@ -159,7 +159,7 @@ describe( 'Shortcode View Constructor', function(){
 		expect( _shortcode.formatShortcode() ).toEqual( '[no_inner_content foo="bar"]burrito[/no_inner_content]' );
 	});
 
-})
+});
 
 }).call(this,typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {})
 },{"../../js/src/models/shortcode":11,"../../js/src/utils/shortcode-view-constructor":12,"../../js/src/utils/sui":13}],5:[function(require,module,exports){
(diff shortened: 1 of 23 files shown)


danielbachhuber:
@goldenapples Can you sanity check this? In particular, 0b7393585cafbd93d9ec1f0cd195eee758194e18 and e0294004737a550abb41213a2119b82f46874fe9

js-tests/build/specs.js
```diff
@@ -281,39 +281,39 @@ describe( "MCE View Constructor", function() {
 	} );
 
 	it( 'parses simple shortcode', function() {
-		var shortcode = MceViewConstructor.parseShortcodeString( '[test_shortcode attr="test value"]')
+		var shortcode = MceViewConstructor.parseShortcodeString( '[test_shortcode attr="test value"]');
```
GaryJones, line 284: Looks like all the highlighted lines here are missing a space before the closing `)`?

js/src/views/edit-attribute-field-attachment.js
```diff
@@ -29,10 +29,6 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
 		if ( editAttributeFieldAttachment.getFromCache( id ) ) {
 			self._renderPreview( editAttributeFieldAttachment.getFromCache( id ) );
 			return;
-
-			// Call the updateValue() function, to trigger any listeners
-			// hooked on it.
-			self.triggerCallbacks();
```
goldenapples, line 35: Nice when linting actually helps you find bugs :)

This shouldn't be removed, but it does need to be moved before the return statement. Otherwise, callback events won't be triggered when changing an attachment to one that's already been previously selected.

goldenapples:
The change in e029400 looks good. 0b73935 actually highlights a bug of mine, which should be fixed rather than removed. 

Overall, this is a big improvement in code quality. I still see a number of examples of `==` - is it worth doing a grep through the plugin for them?

js/build/shortcode-ui.js
```diff
@@ -1273,7 +1272,7 @@ var EditShortcodeForm = wp.Backbone.View.extend({
 
 		} );
 
-		if ( 0 == this.model.get( 'attrs' ).length && ( ! innerContent || typeof innerContent == 'undefined' ) ) {
+		if ( 0 === this.model.get( 'attrs' ).length && ( ! innerContent || typeof innerContent == 'undefined' ) ) {
```
goldenapples, line 1279: The `typeof` comparison here should probably also use strict equality. Also, seems like general practice to check that the typeof an object isn't undefined before accessing the object.

js/src/views/edit-attribute-field-attachment.js
goldenapples, line 35: The bug is mine from c23e329.

danielbachhuber, line 35: Cool, fixing in 3df0e8a

danielbachhuber:
> Overall, this is a big improvement in code quality. I still see a number of examples of == - is it worth doing a grep through the plugin for them?

Eh, separate pull request maybe. I don't know that I'd want to make this change without more test coverage because we might accidentally break something.

PR #474, opened 2015-09-28
Persist unregistered shortcode attributes when editing in TinyMCE

### LINKED ISSUE #425: Persist unregistered shortcode attributes when editing
Currently, if an attribute doesn't have registered UI, it will get lost when the shortcode is edited.

We should persist the data associated with unregistered shortcode attributes.

From https://github.com/fusioneng/Shortcake/issues/401#issuecomment-130729722

danielbachhuber:
Fixes #425


danielbachhuber:
@goldenapples #reviewmerge

I suspect this code is going to change a bit more with #392, but this pull request solves the immediate issue.

goldenapples:
:100:

Issue #477, opened 2015-09-30
Clean up icon vertical alignment in Insert Element UI


isGabe:
I've been working on icon consistency in Shortcake Bakery, and noticed some improvements that could be made in how the icons are positioned in the UI. See https://github.com/fusioneng/shortcake-bakery/issues/95#issuecomment-144104258

danielbachhuber:
:ship:

PR #478, opened 2015-09-30
Update icon positioning, height of icon container and title

isGabe:
See #477


isGabe:
This is what these changes would look like now, before landing the icon consistent work being done in https://github.com/fusioneng/shortcake-bakery/issues/95

screenshot 2015-09-30 17 11 50

danielbachhuber:
@FStop looks pretty solid. ready for it to land?

isGabe:
@danielbachhuber yep, I think it's good to go. #reviewmerge

khromov:
@FStop @danielbachhuber Looks like this PR hides text overflow if it's more than one line. Example:

screen shot 2015-10-01 at 17 09 20

screen shot 2015-10-01 at 17 09 17

danielbachhuber:
> Looks like this PR hides text overflow if it's more than one line.

What if we did horizontal scroll instead?

khromov:
@danielbachhuber Horizontal scroll inside each icon? In that case I'd prefer that the "name badge" grew form the bottom and upwards like it did before, even if that means that the icon won't be perfectly positioned.

isGabe:
@khromov @danielbachhuber Hmm yep, probably a result of adding a `height` to that element. We could probably do `min-height` instead...I can  open a new PR taking multiline text into account later today.

Horizontal scrolling is an interesting idea, but scrollbars on an element that small would probably be brutal.

khromov:
Sounds like a good idea to add `min-height` @FStop !

PR #479, opened 2015-10-01
Add CSS utility classes to fields

khromov:
It's currently almost impossible to add your own CSS to change the appearance of specific shortcodes and fields in the Shortcake UI modal.

This pull request adds a class to the outer body that provides the shortcode base which is currently being edited, as well as a class on each form field that displays the current attribute.

Open to changing the naming convention and any other feedback.

**Before**
screen shot 2015-10-01 at 16 50 20

**After**
screen shot 2015-10-01 at 16 49 47

@danielbachhuber @goldenapples

### DIFF
--- a/inc/fields/class-field-attachment.php
+++ b/inc/fields/class-field-attachment.php
@@ -90,7 +90,7 @@ class Shortcake_Field_Attachment {
 		?>
 
 		<script type="text/html" id="tmpl-fusion-shortcake-field-attachment">
-			<div class="field-block ui-field-attachment ui-attribute-{{ data.attr }}">
+			<div class="field-block shortcode-ui-field-attachment shortcode-ui-attribute-{{ data.attr }}">
 				<label for="{{ data.attr }}">{{ data.label }}</label>
 				<div class="shortcake-attachment-preview attachment-preview attachment">
 					<button id="{{ data.attr }}" class="button button-small add">{{ data.addButton }}</button>
--- a/inc/fields/class-field-attachment.php
+++ b/inc/fields/class-field-attachment.php
@@ -90,7 +90,7 @@ class Shortcake_Field_Attachment {
 		?>
 
 		<script type="text/html" id="tmpl-fusion-shortcake-field-attachment">
-			<div class="field-block">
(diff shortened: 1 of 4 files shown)


inc/fields/class-field-attachment.php
```diff
@@ -90,7 +90,7 @@ public function action_shortcode_ui_loaded_editor() {
 		?>
 
 		<script type="text/html" id="tmpl-fusion-shortcake-field-attachment">
-			<div class="field-block">
+			<div class="field-block ui-field-attachment ui-attribute-{{ data.attr }}">
```
danielbachhuber: Can we use `shortcode-ui` as the prefix, instead of `ui`, for consistency?

inc/fields/class-field-color.php
```diff
@@ -109,7 +109,7 @@ public function load_template() {
 		?>
 
 		<script type="text/html" id="tmpl-fusion-shortcake-field-color">
-			<div class="field-block">
+			<div class="field-block ui-field-color ui-attribute-{{ data.attr }}">
```
danielbachhuber: Same as above.

inc/templates/edit-form.tpl.php
```diff
@@ -2,7 +2,7 @@
 	<form class="edit-shortcode-form">
 		<p><a href="#" class="edit-shortcode-form-cancel">&#8592; <?php esc_html_e( 'Back to list', 'shortcode-ui' ); ?></a></p>
 
-		<div class="edit-shortcode-form-fields"></div>
+		<div class="edit-shortcode-form-fields shortcake-edit-{{ data.model.attributes.shortcode_tag }}"></div>
```
danielbachhuber: Prefix should be `shortcode-ui`, not `shortcake`

inc/fields/class-field-attachment.php
khromov: Sure thing!  047a743

goldenapples:
Looks good to me!

danielbachhuber:
Thanks @khromov !

Issue #371, opened 2015-07-01
Implement batch fetching of shortcode previews


goldenapples:
Currently Shortcake makes a separate ajax request for each shortcode in a post content. It would speed up the initial time it takes to get an editor instance to a stable state if all of these requests could be combined into one request on the initial load.

I haven't investigated too deeply to determine the level of complexity for this, but if it's possible, it could make the experience of opening a post with many shortcodes much smoother.

danielbachhuber:
:100:

bobbingwide:
Is this a bit like running do_shortcode on $post_content with an intercepting function to wrap the output of each shortcode with whatever the editor needs to detect the expanded shortcodes?

goldenapples:
> Is this a bit like running do_shortcode on $post_content with an intercepting function to wrap the output of each shortcode with whatever the editor needs to detect the expanded shortcodes?

In theory, yes. Functionally, it'll work a lot different because the content is being fetched through `wp.mce.views`. Currently each view has its own `fetch()` method which makes an ajax post to `do_shortcode`, and renders the response. I'm picturing passing all the shortcodes to render off to another class which makes a single call for all the views, which will then return a promise that each view can handle on its own.

danielbachhuber:
:ship:

Issue #372, opened 2015-07-01
Add additional attributes to the ajax `do_shortcode` response


goldenapples:
We could add a way for shortcodes to return some attributes like suggested dimensions, float value, or basic css styles, along with the shortcode markup. 

This would involve a few steps:
- building our own ajax endpoint, which would wrap `do_shortcode` and augment it with additional functionality available in shortcodes which were registered with Shortcake.
- Adding additional arguments which could be registered with a shortcode to provide additional information. I imagine these arguments living in a register function that lives in parallel to `shortcode_ui_register_for_shortcode`, maybe something like `shortcode_preview_helpers_register_for_shortcode`. Developers who wanted to offer additional layout tips for their shortcodes could do that there by registering a function or static value to be returned for one or more attribute fields:
  - class names to be applied to the shortcode preview body.
  - initial dimensions to render the preview at
  - inline styles to attach to the preview, ie `display: inline;`

mkeppel:
This morning we got a report of slow List content type that is heavy with Post element gifs. 

Convo with Dodai: https://fusionnetwork.slack.com/archives/edit-tech/p1435842468002515 

Granted this post will have 50+ list items each with a gif in them but the editor is experiencing extremely slow scrolling speeds on the page. Just noting here so that we can have this tracked I have not gotten many requests of slow WP editor experiences but will keep an ear out for them now. 

@goldenapples and I chatted about the situation and potentials for improvements in slack here: https://fusionnetwork.slack.com/archives/tech/p1435842625005052

goldenapples:
It strikes me that this would be a way of solving several rendering-related issues, including #407 and #372.

PR #480, opened 2015-10-01
Add "Fetcher" class for bulk fetching of shortcode previews

goldenapples:
Adds a `Fetcher` utility class to combine and bulk-send requests for
shortcode previews. Currently - until I can find some reliable events on
MCE processing to hook to - this is being done in a very naive way:
every time a request is made, it sets a half-second timeout, so that any
other requests needed can be gathered and sent together.

See #371

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -68,6 +68,7 @@ class Shortcode_UI {
 		add_action( 'admin_enqueue_scripts', array( $this, 'action_admin_enqueue_scripts' ) );
 		add_action( 'wp_enqueue_editor',     array( $this, 'action_wp_enqueue_editor' ) );
 		add_action( 'wp_ajax_do_shortcode',  array( $this, 'handle_ajax_do_shortcode' ) );
+		add_action( 'wp_ajax_bulk_do_shortcode',  array( $this, 'handle_ajax_bulk_do_shortcode' ) );
 		add_filter( 'wp_editor_settings',    array( $this, 'filter_wp_editor_settings' ), 10, 2 );
 	}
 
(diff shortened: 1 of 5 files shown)


khromov:
Nice!

Initial load works fine but if I edit an existing shortcode and press the "Update" button, it gets stuck on the loading screen. Tested on Firefox 41.

goldenapples:
> Initial load works fine but if I edit an existing shortcode and press the "Update" button, it gets stuck on the loading screen. 

Ah thanks. that was something I meant to get around to refactoring but didn't fully get working. I'm trying to use the old `fetch()` method on updating a shortcode, and the new `delayedFetch()` to bulk-render on initial load. Needs some more work apparently...

goldenapples:
I'm thinking, because the Ajax response is now coming back as an array, this will make #372 a lot easier and open us up to handle alignment, non-block-level elements, and so on.

goldenapples:
@danielbachhuber review/merge?

inc/class-shortcode-ui.php
```diff
@@ -338,8 +324,66 @@ public function handle_ajax_do_shortcode() {
...
+					'query' => $posted_query,
+					'response' => $this->render_shortcode_for_preview( $shortcode, $post_id ),
+				);
+			}
+
+			wp_send_json_success( $responses );
+			exit;
+		}
+
+	}
```
danielbachhuber, line 362: Why do we need two separate methods? Can't we always just use one method which always accepts batch requests?

js/src/utils/fetcher.js
```diff
@@ -0,0 +1,75 @@
...
+
+		fetcher.queries.push({
+			promise: fetchPromise,
+			query: query,
+			counter: query.counter
+		});
+
+		if ( ! fetcher.timeout ) {
+			fetcher.timeout = setTimeout(
+				fetcher.fetchAll, 200
```
danielbachhuber: Is 200 going to be a magic number?

danielbachhuber:
> Currently - until I can find some reliable events on
> MCE processing to hook to - this is being done in a very naive way:
> every time a request is made, it sets a half-second timeout, so that any
> other requests needed can be gathered and sent together.

I think we should wait on figuring out the MCE processing events — timeout approach seems like it's going to be a hack that causes problems later.

goldenapples:
> timeout approach seems like it's going to be a hack that causes problems later

I'm actually kind of thinking of it now as a feature, not a hack. Because all the `Fetcher` object is doing is delaying all `do_shortcode` requests and batching them together, it's able to combine requests from multiple MCE instances. I tested it just now with a list post with 7 shortcodes across 4 different WP_Editor instances, and it handled batching them together perfectly.

Because the timeout is cleared whenever a request is sent, the worst case scenario is that, if an editor takes a ridiculous amount of time to initialize, some of its shortcodes will be requested in a separate batch from the others.

As a matter of fact, the shortcode requests were all batched together even when I reduced the timeout to zero - it seems like the mere fact of putting the fetch request in a timeout and delaying until the current executing script finished was enough to ensure that the all of the shortcode views were added to the request before sending it.

inc/class-shortcode-ui.php
goldenapples, line 362: I left the old method in there because I thought that it was handling something that core sent or expected. I guess not: removed in 87e7623.

danielbachhuber:
:100:

khromov:
@danielbachhuber Just as feedback, it'd be great if the branches could linger for a while before being deleted because otherwise it makes it harder to merge it into our fork. :smile:

danielbachhuber:
> because otherwise it makes it harder to merge it into our fork.

Why don't you just use `master`?

Issue #481, opened 2015-10-01
Icons misaligned in Safari


isGabe:
Looks like Safari is throwing a fit because [we didn't prefix `transform: translate(-50%, -50%)`](https://github.com/fusioneng/Shortcake/blob/e11c29f3051e29267a8129510f8d179bac058ee5/css/shortcode-ui.css#L32). 

<img width="1214" alt="screenshot 2015-10-01 15 04 38" src="https://cloud.githubusercontent.com/assets/1437322/10232436/62e6e272-684e-11e5-8ef6-245617a06b18.png">

I'm so used to working with Autoprefixer that it didn't even register when I was creating those styles in #478. Thankfully Safari will be dropping prefixes for that and many other things in v9. In the meantime, though, this should be fixed.

goldenapples:
:stuck_out_tongue_closed_eyes:

danielbachhuber:
:hammer:

PR #483, opened 2015-10-01
Add PostCSS + Autoprefixer to Grunt build

isGabe:
See #481 

This will give us CSS vendor prefixes when needed, automagically. `npm install` will be needed to install PostCSS and Autoprefixer.

I tried copy/pasta what we have in the Fusion theme Gruntfile/package.json, which is grunt-postcss + autoprefixer-core. I got `autoprefixer-core was deprecated. Use autoprefixer package.`, and sure enough it [says as much on their README](https://github.com/ai/autoprefixer-core).

PostCSS wanted to change the CSS sourcemaps a little bit. Not sure why, but it doesn't seem like anything drastic.

@goldenapples #reviewmerge


isGabe:
Confirmed fix in Safari:

screenshot 2015-10-01 15 47 17

goldenapples:
> I tried copy/pasta what we have in the Fusion theme Gruntfile/package.json, which is grunt-postcss + autoprefixer-core.

Ah, good catch. Can you open a pull there to fix?

PR #486, opened 2015-10-02
Restore capability check when rendering a shortcode preview

danielbachhuber:
Better safe than sorry

Removed in #480


danielbachhuber:
@goldenapples #reviewmerge

goldenapples:
Ah, yeah. Good call.

PR #487, opened 2015-10-02
Use core's regex for parsing shortcodes

danielbachhuber:
See #392

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .DS_Store
 composer.lock
 node_modules
-vendor
+/vendor
 npm-debug.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,9 @@ matrix:
     - php: 5.6
       env: WP_VERSION=latest WP_MULTISITE=1 WP_TRAVISCI=travis:phpunit
     - php: 5.6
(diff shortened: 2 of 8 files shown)


danielbachhuber:
@goldenapples @nikolov-tmw Can you look at the JS test failures and let me know what you think? It seems like core's regex provides different behavior than we want.

goldenapples:
> It seems like core's regex provides different behavior than we want.

Looks to me like the problem in tests is in the memoization of the regex in core; that the return value of `wp.shortcode.regexp` is cached from the first time it's run in the test suite, and so later tests that use shortcode tags which weren't defined at first fail.

If you try unwrapping the `_.memoize` call from around the body of `wp.shortcode.regexp`, everything passes... **except** the "parses shortcode with unquoted attributes" test, which is apparently not supported in core.

danielbachhuber:
> Looks to me like the problem in tests is in the memoization of the regex in core; that the return value of wp.shortcode.regexp is cached from the first time it's run in the test suite, and so later tests that use shortcode tags which weren't defined at first fail.

Huh. Ideas on how we can work around it?

danielbachhuber:
@goldenapples #reviewmerge

PR #490, opened 2015-10-02
Un-memoize the wp-shortcode regexp for tests

goldenapples:
Prevent the shortcode regex from `wp.shortcode.regexp` from being cached by resetting it in betweeen each test in the mceViewConstructor suite.

Merges into #487.

Note that not all tests pass still, because the regex used by core to parse attribute values doesn't support unquoted values. We should decide if this is something that's worth preserving support for, or if its safe to break backwards compatibility in such a major way, before continuing with this approach.

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -190,6 +190,12 @@ var wp = (typeof window !== "undefined" ? window['wp'] : typeof global !== "unde
 
 describe( "MCE View Constructor", function() {
 
+	beforeEach( function() {
+		wp.shortcode.regexp = function( tag ) {
+			return new RegExp( '\\[(\\[?)(' + tag + ')(?![\\w-])([^\\]\\/]*(?:\\/(?!\\])[^\\]\\/]*)*?)(?:(\\/)\\]|\\](?:([^\\[]*(?:\\[(?!\\/\\2\\])[^\\[]*)*)(\\[\\/\\2\\]))?)(\\]?)', 'g' );
+		};
+	});
+
 	sui.shortcodes.push( new Shortcode( {
 		label: 'Test Label',
 		shortcode_tag: 'test_shortcode',
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -337,6 +337,13 @@ describe( "MCE View Constructor", function() {
 	} );
 
 	it( 'parses shortcode with unquoted attributes', function() {
+		var shortcode = MceViewConstructor.parseShortcodeString( '[test_shortcode attr=test]');
+		expect( shortcode instanceof Shortcode ).toEqual( true );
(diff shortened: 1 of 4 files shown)


danielbachhuber:
> Note that not all tests pass still, because the regex used by core to parse attribute values doesn't support unquoted values.

Damn:

```
salty-wordpress ➜  fusion-theme  wp shell
wp> do_shortcode( '[fusion-youtube url=https://www.youtube.com/watch?v=8jzDnsjYv9A]' );
string(160) "<iframe class="shortcake-bakery-responsive" width="640" height="360" src="https://youtube.com/embed/8jzDnsjYv9A?rel=0&#038;showinfo=0" frameborder="0"></iframe>"
```

> We should decide if this is something that's worth preserving support for, or if its safe to break backwards compatibility in such a major way, before continuing with this approach.

If core supports it, we need to support it too.

khromov:
@danielbachhuber But that just means the PHP version of the regex supports it, no? If it  doesn't work in the  `wp.shortcode.regexp`, shouldn't it be filed as a core bug?

goldenapples:
I think the issue is that `wp.shortcode.attrs` is also memoized. I'll update our tests to redefine that before each test as well.

danielbachhuber:
> But that just means the PHP version of the regex supports it, no? If it doesn't work in the wp.shortcode.regexp, shouldn't it be filed as a core bug?

Right, and yes.

js-tests/build/specs.js
```diff
@@ -190,6 +190,12 @@ var wp = (typeof window !== "undefined" ? window['wp'] : typeof global !== "unde
 
 describe( "MCE View Constructor", function() {
 
+	beforeEach( function() {
+		wp.shortcode.regexp = function( tag ) {
+			return new RegExp( '\\[(\\[?)(' + tag + ')(?![\\w-])([^\\]\\/]*(?:\\/(?!\\])[^\\]\\/]*)*?)(?:(\\/)\\]|\\](?:([^\\[]*(?:\\[(?!\\/\\2\\])[^\\[]*)*)(\\[\\/\\2\\]))?)(\\]?)', 'g' );
```
danielbachhuber, line 195: Rather than copy and paste this regex, should our copy in `js-tests/vendor/wp-shortcode.js` be the unmemoized version?

goldenapples:
My mistake, the problem issue wasn't the unquoted attribute, it was the hyphenated attribute name.

A Trac ticket https://core.trac.wordpress.org/ticket/25829 which would have addressed this issue was closed as "invalid", with the explanation that core doesn't support hyphenated attribute values.

danielbachhuber:
Looks like the test was added originally in https://github.com/fusioneng/Shortcake/commit/d8850af948225b8811f5b435a71a93b15dc32b3b#diff-a40aa6eb9f7974846714125d0884d73eR22

Issue #395, opened 2015-07-23
Demo plugin (dev.php) observations


GaryJones:
Just a few thoughts from someone new to Shortcake, and who used the code in `dev.php` as a starting point to add support to a custom plugin. I copied all of the `init` hook callback into my code and amended from there.
- Make it clear that there are actually two examples going on here. Maybe split into two `init` hook callbacks, and comment them as "Minimal" and "Typical"/"Advanced".
- Add more comments. The ones that are there for the configuration keys are definitely useful, but some don't have them. For instance, `inner_content` seems to be a simpler way of adding textarea field without needing all of the `attrs` - that's the effect it had for me anyway - but it has no comment to make it clear. `post_type` was guessable as being the post types that the post element should be available for, but a comment with words to that effect would help clarify it.
- Make it clear somewhere that the Add Media -> Insert Post Element link only appears when there's at least one post element that can be added for the current post type. I was editing a Page, and with `post_type` not commented out, but the Minimal example commented out, it took me a few moments to guess / work out why the link wasn't present and my screen didn't match the screenshots. (Thinking long-term, this may become a UX issue for when it's rolled into Core)
- Internationalise strings. It doesn't matter which text domain is used, but having the labels, buttons, placeholders and other visible strings already wrapped in `__()` will save developers time, and help clarify what is a visible string and what's some other config value.
- The `attrs` array is a strange fellow - in my mind, _most of the keys here are actually for the field that will be displayed in the UI, not for the shortcode attribute_. `type` is always a generic word and it is trickier to work out what it is the type of when thinking in shortcode attributes, but easier if thinking in fields. For me (and I realise how non-BC this would be), it would be clearer to focus on fields and have something like:

``` php
// Available shortcode attributes fields and default values. Required. Array.
// Fields model expects 'shortcode_attr', 'type' and 'label'
// Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date.
'fields' => array(
    array(
        'label'          => __( 'Attachment', 'your-text-domain' ), // Field label
        'type'           => 'attachment', // Field type
        'libraryType'    => array( 'image' ), // Media type to insert
        'addButton'      => __( 'Select Image', 'your-text-domain' ); // Button text that opens Media Library
        'frameTitle'     => __( 'Select Image', 'your-text-domain '); // Media Library frame title
        'shortcode_attr' => 'attachment', // Shortcode attribute this field populates
    ),
    array(
[...]

PR #492, opened 2015-10-05
Example plugin: Split UI setup into separate functions (basic and advanced)

Mte90:
I've added the support for few of the points in that ticket #395


danielbachhuber:
Thanks!

PR #493, opened 2015-10-05
Add additional attributes to the `do_shortcode` ajax response for improved preview rendering

goldenapples:
Adds a way to filter the `wpview-wrap` class names returned by each shortcode for preview in the editor. This is a first step toward allowing floated or inline shortcodes in the editor.

Work in progress on #372

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -292,6 +292,34 @@ class Shortcode_UI {
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


khromov:
Cool initiative!

As a side note, we have implemented floating blocks by just floating the inner container, but it is quite buggy when you try to edit the blocks, so this is most welcome!

## screen shot 2015-10-05 at 17 28 15

screen shot 2015-10-05 at 17 28 18

PR #496, opened 2015-10-06
Permit use of HTML in attributes by encoding them them

mattheu:
I picked up https://github.com/fusioneng/Shortcake/pull/322 - synced up with master, and made a few small changes.

The main change is that only the default value of 'escape' for a single attribute is taken from `shortcodeUIFieldData[ type ]` - this allows you to override on a single attribute basis. EG allowing HTML in a text field - not just textareas. Only textarea is true by default - and doesn't apply if this relates to 'inner_content'

Also added tests.

One question - Should we replace all use of the term 'escape' with 'encode' as we're not really escaping. 

One hitch - Rob orignally wrote some clever code that uses the filter `shortcode_atts_{$shortcode_tag}` to unencode the value automatically. However this relies on a closure and hence is PHP 5.3+ only - and causes tests to fail. This isn't really avoidable - as the shortcode tag isn't available in the callback. I'm thinking of opening a ticket for this - but for now we need to either remove or allow breakage on PHP 5.2.


js-tests/build/specs.js
```diff
@@ -58,7 +59,7 @@ var Shortcode = require('../../js/src/models/shortcode');
 var InnerContent = require('../../js/src/models/inner-content');
 var ShortcodeAttribute = require('../../js/src/models/shortcode-attribute');
 var ShortcodeAttributes = require('../../js/src/collections/shortcode-attributes');
-var $ = (typeof window !== "undefined" ? window['jQuery'] : typeof global !== "undefined" ? global['jQuery'] : null);
+var $ = (typeof window !== "undefined" ? window.jQuery : typeof global !== "undefined" ? global.jQuery : null);
```
danielbachhuber: Can you update your node modules and recompile?

mattheu:
Updated
- Updated all use of 'escape' to 'encoded' for consistency.
- Fixed PHP 5.2 - changed the callback to use `current_filter` to determine the current shortcode from within the callback as reccommended here:  https://core.trac.wordpress.org/ticket/15155#comment:12

danielbachhuber:
> One question - Should we replace all use of the term 'escape' with 'encode' as we're not really escaping.

Yes please.

> Only textarea is true by default - and doesn't apply if this relates to 'inner_content'

For consistency and simplicity, can we just disable it by default for all arguments? I'm concerned with leaving it magically on, as it will always require developer intervention.

mattheu:
Updated - no longer sets any defaults.

inc/class-shortcode-ui.php
```diff
@@ -364,4 +367,40 @@ public function handle_ajax_bulk_do_shortcode() {
...
 	}
+
+	/**
+	 * Decode any attributes
+	 * @param  [type] $out   [description]
+	 * @param  [type] $pairs [description]
+	 * @param  [type] $atts  [description]
+	 * @return [type]        [description]
+	 */
+	function filter_shotcode_atts_decode_encoded( $out, $pairs, $atts ) {
```
danielbachhuber: `filter_shotcode_atts_decode_encoded` should be `filter_shortcode_atts_decode_encoded`

Also, can we mark it explicitly as `public` ?

dev.php
```diff
@@ -61,64 +61,68 @@ function shortcode_ui_dev_advanced_example() {
...
+				array(
+					'label'       => __( 'Attachment', 'your-text-domain' ), // Field label
+					'attr'        => 'attachment', // Field type
+					'type'        => 'attachment',
+					'libraryType' => array( 'image' ), // Media type to insert
+					'addButton'   => __( 'Select Image', 'your-text-domain' ), // Button text that opens Media Library
+					'frameTitle'  => __( 'Select Image', 'your-text-domain ' ), // Media Library frame title
+				),
+				array(
+					'label'   => __( 'Citation Source', 'your-text-domain' ),
```
danielbachhuber: Can we do `esc_html__()` for these translation calls?

inc/class-shortcode-ui.php
danielbachhuber: And, should we add tests for this method?

danielbachhuber:
@goldenapples #reviewmerge

danielbachhuber:
Oh. Should it be `encoded` or `encode` ?

danielbachhuber:
I think I ran into a bug.

danielbachhuber:
> I think I ran into a bug.

If I add `encoded=>true` to an existing shortcode attribute, what's the expected behavior for all of the shortcodes I already have saved to the database?

mattheu:
> Oh. Should it be encoded or encode ?

You're right - should probably be `encode`. I was wondering this as I did it. I'll update.

goldenapples:
> If I add encoded=>true to an existing shortcode attribute, what's the expected behavior for all of the shortcodes I already have saved to the database?

Hmm. It's quite possible for field content which was legitimate before to break on urldecoding / re-encoding. Seems like there needs to be a migration path for this case. I don't know what to suggest that doesn't seem hacky and ugly, but allowing a developer changing a setting to corrupt all the existing data is a pretty big issue.

Could we somehow have a magical attribute which flags which fields have encoded data, that is set on formatting the shortcode? That way rendering the shortcode could read from that value when determining whether to urldecode content?

mattheu:
If you have unencoded data and want to enable encod, this will mostly just work OK. Except if you have added HTML that contains the following characters `"'[]` - in which case your shortcodes are broken already.

Pretty sure there is nothing we can do here as its most likely broken the PHP shortcode parsing too. Actually - side note - perhaps we should 'escape for shortcode attribute' and strip these characters? At least things won't break. [I'll create a separate ticket.](https://github.com/fusioneng/Shortcake/issues/500)

If you have encoded data and want to disable, this will cause you to see encoded data and break things. You will need to run some kind of script. HOWEVER - note that your unencoded string will probably break the shortcode parsing. 

I've had a quick go at writing a CLI command to handle this - however I'm wary that it appears more useful than it really is.

Issue #471, opened 2015-09-28
Line breaks not working properly after editing inner_content


khromov:
Consider a simple shortcode that just prints the inner_content.

After adding it and filling out the following as the text area of inner_content and pressing "Insert element":

![screen shot 2015-09-28 at 20 12 57](https://cloud.githubusercontent.com/assets/1207507/10144347/7ecc0036-661d-11e5-8e48-115f4f178d2b.png)

The result in WYSIWYG is:
![screen shot 2015-09-28 at 20 13 02](https://cloud.githubusercontent.com/assets/1207507/10144358/8f6a33cc-661d-11e5-9fcf-1fde70ffa731.png)

After reloading the page or toggling text/visual mode, the result is:
![screen shot 2015-09-28 at 20 13 08](https://cloud.githubusercontent.com/assets/1207507/10144364/91f34a16-661d-11e5-954f-7f5396ff2677.png)

The difference is that  there are line breaks (`<br>`)  being added when reloading page or toggling visual/text, but not otherwise.

Code:

``` php
function sct_init() {
  add_shortcode( 'simple', 'sct_simple_shortcode' );

  if ( function_exists( 'shortcode_ui_register_for_shortcode' ) ) {
    shortcode_ui_register_for_shortcode(
      'simple',
      array(
        'label'         => 'Plain Content',
        'listItemImage' => 'dashicons-welcome-write-blog',
        'inner_content' => array(
          'label' => 'Content',
        ),
        'attrs'         => array(),
      )
    );
  }

}
add_action( 'init', 'sct_init', 10 );

function sct_simple_shortcode( $atts, $content = null ) {
  extract( shortcode_atts(array(), $atts, 'simple' /* Shortcode name for attribute filtering */ ) );

  return do_shortcode( $content );
}
```

khromov:
Fixed in #498

danielbachhuber:
Spent a little bit of time looking into this. Here's my test case:

Pull quote with multiple line breaks:

![image](https://cloud.githubusercontent.com/assets/36432/10546696/66572ec0-73e4-11e5-9122-b80fd43564cb.png)

Renders without the linebreaks:

![image](https://cloud.githubusercontent.com/assets/36432/10546701/71021dbc-73e4-11e5-89a8-f95154fa8ef0.png)

However, the preview is passed the appropriate data:

[...]

PR #498, opened 2015-10-06
Run autop to avoid multiline inner_content wrapping

khromov:
If you use inner_content, there is a problem where multi-line entries get truncated into a single line after you press "Update" in the Shortcake modal. (initial load and toggling between text/visual work fine)

This pull request uses `autop()` before inserting the content into the editor. We already run `unautop()` before populating the edit modal so this appears to have no ill effects and all tests pass.

2nd opinion, @danielbachhuber @goldenapples ?

Example of problem before the patch:
screencast 2015-10-06 16-30-32

### DIFF
--- a/js/src/models/shortcode.js
+++ b/js/src/models/shortcode.js
@@ -87,6 +87,11 @@ Shortcode = Backbone.Model.extend({
 			content = this.get( 'inner_content_backup' );
 		}
 
+		//Run autop before shooting into view
+		if( content ) {
+			content = window.wp.editor.autop( content );
+		}
+
 		if ( attrs.length > 0 ) {
 			template = "[{{ shortcode }} {{ attributes }}]";
 		} else {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -141,6 +141,7 @@ module.exports = function( grunt ) {
 						'js-tests/vendor/wp-shortcode.js',
 						'js-tests/vendor/wp-util.js',
 						'js-tests/vendor/wp-editors.js',
+						'js-tests/vendor/editor.js',
 						'js-tests/vendor/mock-ajax.js',
 					],
 				}
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -534,6 +534,11 @@ Shortcode = Backbone.Model.extend({
 			content = this.get( 'inner_content_backup' );
 		}
 
+		//Run autop before shooting into view
+		if( content ) {
+			content = window.wp.editor.autop( content );
+		}
+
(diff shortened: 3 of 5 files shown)


js/src/models/shortcode.js
```diff
@@ -87,6 +87,11 @@ Shortcode = Backbone.Model.extend({
 			content = this.get( 'inner_content_backup' );
 		}
 
+		//Run autop before shooting into view
```
danielbachhuber, line 90: Can we have the comment explain _why_ it's doing it? The "what" is more straightforward.

js/src/models/shortcode.js
```diff
@@ -87,6 +87,11 @@ Shortcode = Backbone.Model.extend({
...
+		if( content ) {
+			content = window.wp.editor.autop( content );
```
danielbachhuber, line 92: How does this work when TinyMCE isn't instantiated?

danielbachhuber:
I don't fully grok the problem, so it'd be great to get input from @goldenapples and @mattheu 

> this appears to have no ill effects and all tests pass.

Could you write a test case for this scenario?

khromov:
@danielbachhuber You can see a more thorough example of this bug in #471 and it has more screenshots and code example as well.

From my findings TinyMCE  does run `autop()` when you make an initial load of the page or when you switch between visual/text tabs, and the preview is rendered correctly. However, if you open a Shortcake modal and enter multiple lines into inner_content and press the update button, this step isn't performed and the content gets squished onto a single line until you save the content (in which case PHP runs `wpautop()` and the problem is resolved until the next time you edit the shortcode.

I'll see about adding a test and better safeguards when TinyMCE isn't initialized.

js/src/models/shortcode.js
mattheu, line 92: Elsewhere we're already doing something similar: https://github.com/fusioneng/Shortcake/blob/master/js/src/utils/shortcode-view-constructor.js#L231

Basically creating a custom unAutoP method that internally calls the core WP functionality. It allows us to change this more easily in the future.

Actually - the existing code uses the unAutoP functionality in `switchEditors` I wonder whether we should use this for autoP too. Not sure if the functionliaty is different.

danielbachhuber:
This isn't the right approach to solving the problem, because it adds `<p>` and `<br />` HTML to the inner content. While doing so solves the perceived problem, it's an imprecise solution because modifies data permanently (and uncontrollably).

The proper solution is to persist the expected line break data (see https://core.trac.wordpress.org/ticket/34329), and then to run `wpautop` on `$content` in your shortcode callback. If we implement a workaround, it will need to be to solve the core issue.

Issue #502, opened 2015-10-09
HTML characters in localised strings 


bastienho:
Hi,

I understand that escaping strings is better fir security in:

``` php
wp_localize_script( 'shortcode-ui', ' shortcodeUIData', array(
            'shortcodes'      => $shortcodes,
            'strings'         => array(
                'media_frame_title'                 => esc_html__( 'Insert Post Element', 'shortcode-ui' ),
                'media_frame_menu_insert_label'     => esc_html__( 'Insert Post Element', 'shortcode-ui' ),
                'media_frame_menu_update_label'     => esc_html__( '%s Details', 'shortcode-ui' ), // Substituted in JS
                'media_frame_toolbar_insert_label'  => esc_html__( 'Insert Element', 'shortcode-ui' ),
                'media_frame_toolbar_update_label'  => esc_html__( 'Update', 'shortcode-ui' ),
                'media_frame_no_attributes_message' => esc_html__( 'There are no attributes to configure for this Post Element.', 'shortcode-ui' ),
                'edit_tab_label'                    => esc_html__( 'Edit', 'shortcode-ui' ),
                'mce_view_error'                    => esc_html__( 'Failed to load preview', 'shortcode-ui' ),
                'search_placeholder'                => esc_html__( 'Search', 'shortcode-ui' ),
                'insert_content_label'              => esc_html__( 'Insert Content', 'shortcode-ui' ),
            ),
            'nonces'     => array(
                'preview'        => wp_create_nonce( 'shortcode-ui-preview' ),
                'thumbnailImage' => wp_create_nonce( 'shortcode-ui-get-thumbnail-image' ),
            ),
        ) );
```

but in french, **"Insert Post Element"** is translated has **"Insérer un élément d'article"** wich appears like this:

**Insérer un élément d&#039;article**

wich is pretty ugly...

Can you use `esc_attr` instead ?

danielbachhuber:
Thanks for flagging. Because the JS templates handle escaping for us, we don't need to prematurely escape. This is fixed on master, and will be released in v0.6.0.

Issue #456, opened 2015-09-11
Translation issue with apostrophe character 


btribouillet:
Hi,

It seems there is a problem with the french translation with apostrophe character .

![capture d ecran 2015-09-11 a 14 57 04](https://cloud.githubusercontent.com/assets/7518508/9815369/c2906b44-5895-11e5-802c-5cf1a1d1f692.png)

It should be "d'article"

danielbachhuber:
Hm. Not sure where that's at in the UI. It seems like it's probably getting double-encoded, though.

Can you share a bigger screenshot, or do some debugging on your own?

danielbachhuber:
Related #502

This is now fixed on master.

PR #504, opened 2015-10-09
Prevent double-escaping in JS templates by removing escaping here

danielbachhuber:
See #502, #456

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -223,7 +223,6 @@ class Shortcode_UI {
 				'media_frame_toolbar_insert_label'  => __( 'Insert Element', 'shortcode-ui' ),
 				'media_frame_toolbar_update_label'  => __( 'Update', 'shortcode-ui' ),
 				'media_frame_no_attributes_message' => __( 'There are no attributes to configure for this Post Element.', 'shortcode-ui' ),
-				'edit_tab_label'                    => __( 'Edit', 'shortcode-ui' ),
 				'mce_view_error'                    => __( 'Failed to load preview', 'shortcode-ui' ),
 				'search_placeholder'                => __( 'Search', 'shortcode-ui' ),
 				'insert_content_label'              => __( 'Insert Content', 'shortcode-ui' ),
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -217,16 +217,16 @@ class Shortcode_UI {
 		wp_localize_script( 'shortcode-ui', ' shortcodeUIData', array(
 			'shortcodes'      => $shortcodes,
 			'strings'         => array(
-				'media_frame_title'                 => esc_html__( 'Insert Post Element', 'shortcode-ui' ),
-				'media_frame_menu_insert_label'     => esc_html__( 'Insert Post Element', 'shortcode-ui' ),
-				'media_frame_menu_update_label'     => esc_html__( '%s Details', 'shortcode-ui' ), // Substituted in JS
-				'media_frame_toolbar_insert_label'  => esc_html__( 'Insert Element', 'shortcode-ui' ),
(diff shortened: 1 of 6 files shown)


danielbachhuber:
@goldenapples #reviewmerge

inc/class-shortcode-ui.php
```diff
@@ -217,16 +217,16 @@ public function enqueue() {
...
-				'search_placeholder'                => esc_html__( 'Search', 'shortcode-ui' ),
-				'insert_content_label'              => esc_html__( 'Insert Content', 'shortcode-ui' ),
+				'media_frame_title'                 => __( 'Insert Post Element', 'shortcode-ui' ),
+				'media_frame_menu_insert_label'     => __( 'Insert Post Element', 'shortcode-ui' ),
+				'media_frame_menu_update_label'     => __( '%s Details', 'shortcode-ui' ), // Substituted in JS
+				'media_frame_toolbar_insert_label'  => __( 'Insert Element', 'shortcode-ui' ),
+				'media_frame_toolbar_update_label'  => __( 'Update', 'shortcode-ui' ),
+				'media_frame_no_attributes_message' => __( 'There are no attributes to configure for this Post Element.', 'shortcode-ui' ),
+				'edit_tab_label'                    => __( 'Edit', 'shortcode-ui' ),
+				'mce_view_error'                    => __( 'Failed to load preview', 'shortcode-ui' ),
```
goldenapples, line 226: This [isn't being escaped on output](https://github.com/fusioneng/Shortcake/blob/master/js/src/utils/shortcode-view-constructor.js#L26).

danielbachhuber, line 226: Ok. I'll update so we're properly building those strings.

inc/class-shortcode-ui.php
```diff
@@ -217,16 +217,16 @@ public function enqueue() {
...
+				'search_placeholder'                => __( 'Search', 'shortcode-ui' ),
```
goldenapples, line 227: Neither is [this](https://github.com/fusioneng/Shortcake/blob/master/js/src/views/search-shortcode.js#L15), from what I can tell.

inc/class-shortcode-ui.php
```diff
@@ -217,16 +217,16 @@ public function enqueue() {
...
-				'mce_view_error'                    => esc_html__( 'Failed to load preview', 'shortcode-ui' ),
-				'search_placeholder'                => esc_html__( 'Search', 'shortcode-ui' ),
-				'insert_content_label'              => esc_html__( 'Insert Content', 'shortcode-ui' ),
+				'media_frame_title'                 => __( 'Insert Post Element', 'shortcode-ui' ),
+				'media_frame_menu_insert_label'     => __( 'Insert Post Element', 'shortcode-ui' ),
+				'media_frame_menu_update_label'     => __( '%s Details', 'shortcode-ui' ), // Substituted in JS
+				'media_frame_toolbar_insert_label'  => __( 'Insert Element', 'shortcode-ui' ),
+				'media_frame_toolbar_update_label'  => __( 'Update', 'shortcode-ui' ),
+				'media_frame_no_attributes_message' => __( 'There are no attributes to configure for this Post Element.', 'shortcode-ui' ),
+				'edit_tab_label'                    => __( 'Edit', 'shortcode-ui' ),
```
goldenapples: This is not being used anywhere, AFAICT...

inc/class-shortcode-ui.php
goldenapples, line 227: Never mind, it's being escaped in `wp.media.view.Search`.

goldenapples:
Most of the strings we're using are being properly built using js templates or with `$.text()` or similar. I just saw one cases that wasn't, which should be updated, and one string which isn't being used at all, which can be removed.

danielbachhuber:
@goldenapples ready to :ship:

goldenapples:
Looks good to me.

PR #505, opened 2015-10-09
Use `load_textdomain` instead of `load_plugin_textdomain()`

danielbachhuber:
The latter doesn't work when the plugin is installed in non-standard locations.

### DIFF
--- a/shortcode-ui.php
+++ b/shortcode-ui.php
@@ -50,7 +50,13 @@ function shortcode_ui_init() {
  * @return null
  */
 function shortcode_ui_load_textdomain() {
-	load_plugin_textdomain( 'shortcode-ui', false, dirname( plugin_basename( __FILE__ ) ) . '/languages/' );
+	// Don't use load_plugin_textdomain because it doesn't support non-standard directories
+	// See https://core.trac.wordpress.org/ticket/23794
+	$locale = get_locale();
+	$domain = 'shortcode-ui';
+	$locale = apply_filters( 'plugin_locale', $locale, $domain );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@goldenapples #reviewmerge

shortcode-ui.php
```diff
@@ -50,7 +50,13 @@ function shortcode_ui_init() {
...
  */
 function shortcode_ui_load_textdomain() {
-	load_plugin_textdomain( 'shortcode-ui', false, dirname( plugin_basename( __FILE__ ) ) . '/languages/' );
+	// Don't use load_plugin_textdomain because it doesn't support non-standard directories
+	// See https://core.trac.wordpress.org/ticket/23794
+	$locale = get_locale();
+	$domain = 'shortcode-ui';
+	$locale = apply_filters( 'plugin_locale', $locale, $domain );
+	$mofile = dirname( __FILE__ ) . '/languages/' . $domain . '-' . $locale . '.mo';
+	load_textdomain( $domain, $mofile );
```
goldenapples, line 67: Core has an additional fallback to loading translations from `WP_LANG_DIR`, which is something like `wp-content/languages`. I'm not too familiar with how that's used, but I can imagine that's useful for users maintaining local translations for when plugins don't include their language. Should we include the same behavior?

danielbachhuber, line 67: > I can imagine that's useful for users maintaining local translations for when plugins don't include their language. Should we include the same behavior?

Oh, true. That's probably best. I'll fix this up.

danielbachhuber:
@goldenapples this is good to :ship:

PR #508, opened 2015-10-12
Attachment Gallery

mattheu:
Support multiple: true on the attachments field.

2 issues i've run into.
1. the code to select the attachments if you edit the selection works the first time, but not the second. I cannot work out why not ;)
2. How to display the attachment data stuff.
   - Also could we make the display of this optional?


mattheu:
Oh gotta sync this up.

mattheu:
Updated - think this is ready for a review

css/sass/_field-image.scss
```diff
@@ -1,110 +1,112 @@
...
 		}
+
+		// &.has-attachment  {
+		// 	.button.add {
+		// 		display: none;
+		// 	}
+		// 	.button.remove {
+		// 		display: block;
+		// 	}
+		// }
```
danielbachhuber: Should this be commented out?

ameshkin (commented):
just what i was looking for

mattheu:
Updated. @goldenapples this should be ready to go.,

goldenapples:
Seems workable as an initial step here.

I'm seeing some weird styling on the `.attachment` element in the shortcode modal:

screen shot 2016-11-08 at 8 11 54 am
screen shot 2016-11-08 at 8 13 39 am
... which is coming from this core style, forcing that to 16.66% of its container, when we actually want it to fill the whole container:
screen shot 2016-11-08 at 8 16 01 am

Also, I'm not sure if the first issue is actually worked out, is it? When I edit a shortcode after sending it to the editor once, I don't see the previously selected items as selected. It would probably be good to figure that behavior out before merging this, as otherwise that will cause some editorial headaches.

mattheu:
I've pushed some updates for the styles. 

I'm no longer seeing any issues with the gallery images not loading, although they it does have to fetch the attachment data, so there is a little delay on this, but they should pop in.

js/src/views/edit-attribute-field-attachment.js
```diff
@@ -87,46 +123,37 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
...
 
-		this.$thumbnailDetailsContainer.find( '.filename' ).text( attachment.filename );
-		this.$thumbnailDetailsContainer.find( '.date-formatted' ).text( attachment.dateFormatted );
-		this.$thumbnailDetailsContainer.find( '.size' ).text( attachment.filesizeHumanReadable );
-		this.$thumbnailDetailsContainer.find( '.dimensions' ).text( attachment.height + ' × ' + attachment.width );
-		this.$thumbnailDetailsContainer.find( '.edit-link a' ).attr( "href", attachment.editLink );
-		this.$thumbnailDetailsContainer.toggleClass( 'has-attachment', true );
+		value.sort( function( a, b ) {
+			return a < b;
+		} );
```
goldenapples, line 154: I'm not totally sure about sorting the selected attachments by attachment ID, but as we don't have a sortable implementation here yet, this will do as a first introduction to the feature.

goldenapples:
Very nice :100:

This looks and works a lot better for me. I wasn't able to reproduce the issue I was seeing before where selected images are unselected when opening the media frame to select new images. I'm not sure if it was a bug that you fixed, or if I was just confused about the UI before. (Having to command-select to add a selection seems strange, especially if you're on a page of images that don't show any attachments selected already.)

The code looks good to me. I'm pretty sure that at some point we'll have to support user-defined ordering of attachments in a gallery, but this is a good chunk of code to merge at one time. I can open a new issue for sortable images, and merge this as is.

js/src/views/edit-attribute-field-attachment.js
mattheu: I'll have a think about how to tackle this one.

mattheu:
Yes command select images isn't totally obvious, but at least lets us achive the goal.

I'll have a look at how to improve this further. Perhaps we can bring in more of the core UI for galleries.

PR #511, opened 2015-10-13
Clarify dev plugin comments; wrap strings in translation functions

danielbachhuber:
See #395

### DIFF
--- a/dev.php
+++ b/dev.php
@@ -96,7 +96,9 @@ function shortcode_ui_dev_advanced_example() {
 			),
 			/*
 			 * Register UI for attributes of the shortcode. Optional.
+			 * If no UI is registered for an attribute, then the attribute value will be backed up during editing.
 			 * Each array must include 'attr', 'type', and 'label'.
+			 * 'attr' should be the name of the attribute.
 			 * Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date, post_select, attachment, color.
 			 * Use 'meta' to add arbitrary attributes to the HTML of the field.
 			 * Use 'encode' to encode attribute data. Requires customization to callback to decode.
--- a/dev.php
+++ b/dev.php
@@ -96,7 +96,11 @@ function shortcode_ui_dev_advanced_example() {
 			),
 			/*
 			 * Register UI for attributes of the shortcode. Optional.
-			 * If no UI is registered for an attribute, then the attribute value will be backed up during editing.
+			 *
+			 * If no UI is registered for an attribute, then the attribute will 
+			 * not be editable through Shortcake's UI. However, the value of any 
+			 * unregistered attributes will be preserved when editing.
+			 * 
 			 * Each array must include 'attr', 'type', and 'label'.
 			 * 'attr' should be the name of the attribute.
 			 * Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date, post_select, attachment, color.
--- a/dev.php
+++ b/dev.php
@@ -18,12 +18,14 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  */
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@goldenapples #reviewmerge

dev.php
```diff
@@ -54,43 +60,75 @@ function shortcode_ui_dev_advanced_example() {
...
+				'label'        => esc_html__( 'Quote', 'shortcode-ui' ),
+				'description'  => esc_html__( 'Include a statement from someone famous.', 'shortcode-ui' ),
 			),
-			'post_type' => array( 'post' ), //Post type support
-			// Available shortcode attributes and default values. Required. Array.
-			// Attribute model expects 'attr', 'type' and 'label'
-			// Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date.
+			/*
+			 * Register UI for attributes of the shortcode. Optional.
+			 * Each array must include 'attr', 'type', and 'label'.
```
goldenapples, line 104: One thing which may not be obvious on first glance is that `attr` is the attribute name.

danielbachhuber, line 104: Clarified 75c4226c6dbf60b6a485f82f98d5422e9e4e42c0

Maybe we should rename it to `name`

dev.php
```diff
@@ -54,43 +60,77 @@ function shortcode_ui_dev_advanced_example() {
...
+				'label'        => esc_html__( 'Quote', 'shortcode-ui' ),
+				'description'  => esc_html__( 'Include a statement from someone famous.', 'shortcode-ui' ),
 			),
-			'post_type' => array( 'post' ), //Post type support
-			// Available shortcode attributes and default values. Required. Array.
-			// Attribute model expects 'attr', 'type' and 'label'
-			// Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date.
+			/*
+			 * Register UI for attributes of the shortcode. Optional.
+			 * If no UI is registered for an attribute, then the attribute value will be backed up during editing.
```
goldenapples: Hmm. This doesn't sound intuitive to me. I think it might read better like "If no UI is registered for an attribute, then the attribute will not be editable through Shortcake's UI. However, the value of any unregistered attributes will be preserved when editing."

dev.php
goldenapples, line 104: > Maybe we should rename it to `name`

Yeah, I think that would be clearer.

dev.php
danielbachhuber: 52f3c095772bcea7b467494fd3444165ae6986f3

dev.php
```diff
@@ -54,43 +60,81 @@ function shortcode_ui_dev_advanced_example() {
...
+			/*
+			 * Register UI for attributes of the shortcode. Optional.
+			 *
+			 * If no UI is registered for an attribute, then the attribute will 
+			 * not be editable through Shortcake's UI. However, the value of any 
+			 * unregistered attributes will be preserved when editing.
+			 * 
+			 * Each array must include 'attr', 'type', and 'label'.
+			 * 'attr' should be the name of the attribute.
+			 * Supported field types: text, checkbox, textarea, radio, select, email, url, number, and date, post_select, attachment, color.
```
goldenapples: Can you add line wrapping here? Also, wording like `'type' should be a supported field type: ...` would be more parallel to other explanations here.

goldenapples:
Left one more comment.

Also one thing I'm thinking here is that it would be good to demonstrate using the new hook `register_shortcode_ui` for registering UI on, rather than hooking everything on `init`.

dev.php
danielbachhuber: 743a8277f6afa800ca311fc4e5d162f1dfd70653

danielbachhuber:
> Also one thing I'm thinking here is that it would be good to demonstrate using the new hook register_shortcode_ui for registering UI on, rather than hooking everything on init.

I'll do that in a follow-up PR.

Issue #515, opened 2015-10-16
Backbone events doesn't detect paste properly


r-a-y:
Hi!

Great work on Shortcake!  I've been looking into the Event Attribute callbacks wiki page:
https://github.com/fusioneng/Shortcake/wiki/Event-Attribute-Callbacks

And have started implementing work on this in my shortcode plugin.

One issue I've run into is pasting some text into an input text field to toggle some additional fields.  (To paste, I'm using <kbd>CTRL</kbd>+<kbd>V</kbd>.)

When adding a `console.log( updatedVal )` statement inside the wiki example code, it appears that the paste event is happening too quickly and the attribute value later gets wiped out; thus, not showing my additional fields.  The additional fields only pop up once I type in an additional character to the input text field.

I tried to remedy this by adding a `setTimeout` in the `inputChanged()` method:
https://github.com/fusioneng/Shortcake/blob/f077eb7a85bfa80985717add3f06f66f5bc878ec/js/build/shortcode-ui.js#L1124

But that didn't seem to work properly.  Let me know if you need me to clarify anything.

goldenapples:
Hmm. I think that has to do with [the events we're binding to](https://github.com/fusioneng/Shortcake/blob/master/js/src/views/edit-attribute-field.js#L9-L20) to listen to a change in the model.

Specifically, we're listening to `keyup`, which happens when you release the "V", and not necessarily after the pasted text is inserted into the input box.

I wonder if changing our events on editAttributeField to listen for `input` instead would work better.

r-a-y:
Thanks for the quick reply, goldenapples.

Changing:
`'keyup input[type="text"]':   'inputChanged',`

to:
`'input input[type="text"]':   'inputChanged',`

Works for me.

goldenapples:
> Changing: `'keyup input[type="text"]'` to `'input input[type="text"]'` works for me.

Great. I can't think of any cases offhand where it wouldn't work. I'll open a pull request with that change and see if it works more generally to change from keyup to input.

r-a-y:
Fixed in #516.

Thanks @goldenapples!

PR #516, opened 2015-10-16
Bind to `input` event rather than `keyup`

goldenapples:
For all attribute fields whose type is `input` or `textarea`, binding to
the `input` event gives a much more reliable indicator of when the
field's value has changed. This is fired every time the input's value
changes - including paste events, rather than only when it's changed by
actually typing into the field.

See #515 for an example of the behaviour this is intended to fix.


goldenapples:
I think this is smoother and more expected behavior for these fields. @danielbachhuber @mattheu What do you think?

danielbachhuber:
:shipit:

Issue #485, opened 2015-10-02
Textarea does not support linebreaks in parameters


niabot:
If i use a textarea for input into a _normal parameter_ (not _inner_content_) then the linebreaks will give strange results. Once i safe the shortcode from the ui, reload the page and try editing again, the content of the textarea field only shows the first line with a `</p>` appended to it.

The shortcut itself looks like this inside the editor (not WYSIWYG):

```
[myshortcode param="Some text

More text" param2="xyz"]
```

Result in textarea:

```
Some Text</p>
```

khromov:
Should be fixable by the changeset in #496

danielbachhuber:
Pretty sure this is caused by https://core.trac.wordpress.org/ticket/34329

Related #471

PR #518, opened 2015-10-19
Reverse effects of wpautop when accepting shortcode data from TinyMCE views

### LINKED ISSUE #471: Line breaks not working properly after editing inner_content

danielbachhuber:
Because TinyMCE views are rendered _after_ wpautop is applied, we need to reverse the effect before Shortcake starts handling the shortcode.

Fixes #471, #485

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -824,7 +824,8 @@ var shortcodeViewConstructor = {
 	 * @param {Object} options Options
 	 */
 	getShortcodeModel: function( options ) {
-		var shortcodeModel;
+		var shortcodeModel,
+			self = this;
 
 		shortcodeModel = sui.shortcodes.findWhere( { shortcode_tag: options.tag } );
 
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@goldenapples #reviewmerge

It still feels dirty to strip these tags, but I guess if that's what core does anyway, we should do it too.

goldenapples:
I think this makes sense. I don't particularly like it either, but this should do no more damage to post content than switching back and forth between the visual and text editor will do.

PR #519, opened 2015-10-19
Update grunt sass

mattheu:
I was getting some errors locally when compilling the CSS.

> Error: `libsass` bindings not found. Try reinstalling `node-sass`?

I think its related to updating node +/ npm. I uninstalled grunt-sass and reinstalled at latest and it fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "grunt-contrib-watch": "^0.6.1",
     "grunt-phpcs": "^0.4.0",
     "grunt-postcss": "^0.6.0",
-    "grunt-sass": "^1.0.0",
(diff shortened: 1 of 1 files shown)


khromov:
You are correct, `libsass` appears to be compiled to a specific version of node.

I'd suggest also adding NPM shrinkwrap to avoid people working with different versions of the packages.
https://docs.npmjs.com/cli/shrinkwrap

mattheu:
Updated to latest version.

Shrinkwrap looks cool, but I think we should tackle separately. This needs updating - so I'm going to go ahead and merge.

PR #521, opened 2015-10-25
Added Fronted support

jesperbjerke:


danielbachhuber:
See https://github.com/fusioneng/Shortcake/pull/522#issuecomment-150953847

PR #522, opened 2015-10-25
Added Fronted support

jesperbjerke:


danielbachhuber:
Sorry, because we can't support compatibility with all plugins, we can't add a plugin compatibility shim for Fronted. You'll want to include the equivalent code in your plugin.

PR #529, opened 2015-10-28
Improve demo file

### LINKED ISSUE #395: Demo plugin (dev.php) observations

GaryJones:
See individual commits for details.

Fixes #395.

### DIFF
--- a/dev.php
+++ b/dev.php
@@ -135,7 +135,8 @@ function shortcode_ui_dev_advanced_example() {
 	/*
 	 * Define the UI for attributes of the shortcode. Optional.
 	 *
-	 * In this demo example, we register three fields - Attachment, Citation Source, and Select Page.
+	 * In this demo example, we register multiple fields related to showing a quotation
+	 * - Attachment, Citation Source, Select Page, Background Color, Alignment and Year.
 	 *
 	 * If no UI is registered for an attribute, then the attribute will
 	 * not be editable through Shortcake's UI. However, the value of any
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks! Fixing the PHPCS errors in #530

Issue #520, opened 2015-10-22
Shortcode attributes not passed through to AJAX functions


drewbaker:
If I have a short code like this:

```
[product-image-block left-attachment-id="166" product-attachment-id="408" product-url="example.com" product-brand="Kinfolk" product-name="Spotted Shirt" product-price="$99"]
     <h2>December 50% Off</h2>
         Free shipping on everything!
     <a href="http://example.com">View sale item</a>
[/product-image-block]
```

It fires off this AJAX request:
![screen shot 2015-10-22 at 1 50 09 pm](https://cloud.githubusercontent.com/assets/475015/10677932/1b5e2af6-78c4-11e5-82d8-920991aaee87.png)

Notice that the request is missing all the parameters, so the HTML that returns is wrong obviously too.

Does anyone else have this problem? Is there anything that could conflict with that?

khromov:
The shortcode content is sent, and the parameters are parsed on the backend. (Just like WP Core does it.)

drewbaker:
@khromov It doesn't look like the attributes get sent though. Like `left-attachment-id="166" product-attachment-id="408"` isn't included in the POST request?

It might be due to bug #495 though.

goldenapples:
> It might be due to bug #495 though.

Yup, that seems to be the issue here. Can you reproduce with any non-hyphenated attributes?

drewbaker:
Actually, no it doesn't work even without hyphens.

This short code `[product id="1233"]Test[/product]` produces this POST request:

![screen shot 2015-10-22 at 2 19 19 pm](https://cloud.githubusercontent.com/assets/475015/10678558/10149de8-78c8-11e5-8de6-bca30616a149.png)

drewbaker:
In case it's relevant, the site also has WooCommerce installed on it, but otherwise no other plugins.

goldenapples:
hmm. That is strange. What version of Shortcake are you running? The current master has some slightly different behavior (it would be calling `action=bulk_do_shortcode`, rather than `action=do_shortcode`, for example.)

drewbaker:
Version 0.5.0 from the WordPress.org plugin app store (or whatever they call that).

goldenapples:
That clearly seems like a bug, but before we get too deep into tracking it down, can you try downloading the current master version from [here](https://github.com/fusioneng/Shortcake/archive/master.zip) and see if you still get the same issue?

drewbaker:
Ok, trying now...

[...]

PR #531, opened 2015-10-29
Always persist unregistered shortcode attributes and inner content

danielbachhuber:
Even when the shortcode has already been parsed

See #520

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -198,6 +198,23 @@ describe( 'Shortcode View Constructor', function(){
 		var shortcode = ShortcodeViewConstructor.parseShortcodeString( '[no_custom_attribute foo="bar" bar="banana"]' );
 		var _shortcode = $.extend( true, {}, shortcode );
 		expect( _shortcode.formatShortcode() ).toEqual( '[no_custom_attribute foo="bar" bar="banana"]' );
+		ShortcodeViewConstructor.shortcode = {
+			'type' : 'single',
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@goldenapples #reviewmerge

js-tests/build/specs.js
```diff
@@ -180,6 +180,23 @@ describe( 'Shortcode View Constructor', function(){
...
+					'foo' : 'bar',
+				},
+				'numeric' : [],
+			},
+			'content' : 'burrito'
+		};
+		var ShortcodeViewConstructorWithoutFetch = ShortcodeViewConstructor;
+		ShortcodeViewConstructorWithoutFetch.delayedFetch = function() {
+			return new $.Deferred();
+		};
```
goldenapples, line 197: Why the extra variable assignment here? There's nothing wrong with this, it just threw me for a minute reading it.

```
        ShortcodeViewConstructor.delayedFetch = function() {
            return new $.Deferred();
        };
```

would do the same thing, wouldn't it?

goldenapples:
Looks great! I appreciate being able to see how the test coverage works here. That'll help me understand how to build out good coverage for the attribute fields.

I can merge once the build passes.

js-tests/build/specs.js
danielbachhuber, line 197: > Why the extra variable assignment here?

Because I don't want to modify the original? Although, I see below that I'm still using the original in the test...

PR #532, opened 2015-10-29
Don't overload `delayedFetch`

danielbachhuber:
From https://github.com/wp-shortcake/shortcake/pull/531#discussion_r43408981

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -191,10 +191,6 @@ describe( 'Shortcode View Constructor', function(){
 			},
 			'content' : 'burrito'
 		};
-		var ShortcodeViewConstructorWithoutFetch = ShortcodeViewConstructor;
-		ShortcodeViewConstructorWithoutFetch.delayedFetch = function() {
-			return new $.Deferred();
-		};
 		ShortcodeViewConstructor.initialize();
 		expect( ShortcodeViewConstructor.shortcodeModel.formatShortcode() ).toEqual( '[no_inner_content foo="bar"]burrito[/no_inner_content]' );
 	});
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@goldenapples this is the error I got when I don't overload `delayedFetch`. I don't fully grok what's happening though. Thoughts?

goldenapples:
Right, you'll have to overload `delayedFetch` for the tests, because otherwise initializing a shortcode will call [`fetcher.queueToFetch()`](https://github.com/wp-shortcake/shortcake/blob/master/js/src/utils/fetcher.js#L50-L66), which will call [`fetcher.fetchAll()`](https://github.com/wp-shortcake/shortcake/blob/master/js/src/utils/fetcher.js#L77-L101) - which is where the error you're seeing is coming from, as it uses the global `ajaxurl`, which isn't defined in our test suite.

goldenapples:
Why isn't jasmine exiting with an error code when it fails?

danielbachhuber:
> Right, you'll have to overload delayedFetch for the tests

Should we always overload, or is how we have it on `master` fine?

danielbachhuber:
> Why isn't jasmine exiting with an error code when it fails?

Because it only throws an error — the test still passes.

goldenapples:
Also, I notice that I don't have any tests in place for `fetcher`. I'd like to try and write some solid test coverage for it.

Overloading that function for the ShortcodeViewConstructor tests seems fine for now. I imagine we'll want to mock up a response value using jasmine-spy or similar to actually have the internals of our utils classes fully covered.

danielbachhuber:
:+1:

PR #533, opened 2015-10-30
Update README Links

willthemoor:
The links at the top of the README were pointing to pages (wiki, issues) on `/fusioneng/x` that don't exist. eg:

> Once you've done so, you'll need to [register the UI for your code](https://github.com/fusioneng/Shortcake/wiki/Registering-Shortcode-UI).

This just updates them to point at `/wp-shortcake/x`.


danielbachhuber:
Thanks for the pull request! These links actually need to be updated in `readme.txt`, from which `README.md` is generated. I'll handle this in #534

willthemoor:
Ah, thanks for this update. 1) Sorry I updated the wrong file, 2) I had no idea about wp_readme_to_markdown! So double thanks. :)

PR #534, opened 2015-10-30
Update links in the readme to their new URLs

danielbachhuber:
Fixes #533


goldenapples:
I believe Github URLs are case-sensitive, so we'll also have to change `Shortcake` to `shortcake` in those links.

danielbachhuber:
> I believe Github URLs are case-sensitive

Nope https://github.com/wp-shortcake/Shortcake/wiki/Registering-Shortcode-UI

danielbachhuber:
I'll update though

goldenapples:
Ah, cool to know. I could swear I just tried to visit one of the links and got a 404. Must've mistyped it.

Issue #538, opened 2015-11-03
Javascript TypeError when clicking media frame menu items


goldenapples:
From https://github.com/wp-shortcake/image-shortcake/issues/59:

> Full error message (with unnecessary path removed):
> 
> ```
> Uncaught TypeError: Cannot read property 'get' of undefined ... path/to/shortcake/js/build/shortcode-ui.js?ver=0.7.0-alpha line 1608
> ```
> 
> This only happens with the image shortcode, not other shortcodes I've added. It also only happens for newly-uploaded images. If I add the image to editor from the library again after reloading the page, the error does not occur.
> 
> The image appears in the editor despite the error.

---

It looks like Shortcake is listening to clicks on [all links in the media frame menu](https://github.com/wp-shortcake/shortcake/blob/master/js/src/views/media-frame.js#L50) and trying to read the current shortcode information from `this.state().props`. The `props` object, however, is undefined when the click is on one of core's links like "Insert Media" or "Featured Image".

It would be good to check `typeof this.state().props` before trying to get properties from it.

danielbachhuber:
I'll assign this to myself as a note to release v0.6.1 sometime next week.

danielbachhuber:
:ship: :ship:

PR #540, opened 2015-11-04
Check that mediaFrame.state().props is defined before accessing it

goldenapples:
Avoid javascript TypeError when ckicking on a media frame menu item
which doesn't set the props object that Shortcake uses to pass
`currentShortcode` in between modal frames.

See #538


goldenapples:

danielbachhuber:
Huh. I could've sworn we've fixed this before.

Issue #537, opened 2015-11-02
Turkish translation ready >


sinanisler:
Turkish translation ready >

http://www.sinanisler.com/blog/wp-content/uploads/shortcode-ui-tr_TR.zip

danielbachhuber:
Thanks!

sinanisler:
your welcome sir 
good job keep up.

PR #541, opened 2015-11-04
Add Turkish localization files

goldenapples:
Adds `tr_TR` localization files, props @sinanisler

See #537


goldenapples:

PR #542, opened 2015-11-04
Remove executable bit from Turkish language files

goldenapples:
Translation files added in #541 had permissions which allowed
execution. D'oh. Updated them with `chmod a-x` to avoid any possible
evil usage.


goldenapples:
@danielbachhuber #reviewmerge

Not my day for catching problems in review, I guess.

Issue #545, opened 2015-11-11
post_select broken?


jdmoliner:
Hi,

I can not run post_select with WP 4.3.1 and  shortcake 0.6.1. 

It should work? or am I doing something wrong? I get always 'No results were found'

danielbachhuber:
Can you share how you've defined your shortcode UI, and the query you're trying to make?

jdmoliner:
I used the same example as in dev.php:

<pre>
 function shortcode_ui_dev_advanced_example() { 
    shortcode_ui_register_for_shortcode( 'shortcake_dev',
        array(          
            'label' => esc_html__( 'Botón', 'shortcode-ui' ),          
            'listItemImage' => 'dashicons-editor-quote',            
            'attrs' => array(   
                array(
                    'label' => esc_html__( 'Select Page', 'shortcode-ui' ),
                    'attr' => 'page',
                    'type' => 'post_select',
                    'query' => array( 'post_type' => 'page' ),
                    'multiple' => true,
                )
            )
        )
    );
 }
 add_action( 'register_shortcode_ui', 'shortcode_ui_dev_advanced_example' );
</pre>

danielbachhuber:
I've confirmed the bug :( We'll get this fixed in v0.6.2.

For now, you can address the bug locally by registering your UI on `init`, and adding a `function_exists()` check for `shortcode_ui_register_for_shortcode`

danielbachhuber:
@jdmoliner Just released v0.6.2 if you care to upgrade instead.

widoz:
Installed the 0.6.2 but the issue remains.

This is what I seen on console.

> Uncaught Error: No select2/compat/initSelection

May be a different issue?

widoz:
Sorry, found another thread with the same issue of mine.

PR #547, opened 2015-11-12
Ensure `register_shortcode_ui` is always run before getting UI args

danielbachhuber:
Because we need to check shortcode UI args in some non-editor contexts,
we need to make sure the `register_shortcode_ui` hook is always run.

Specifically, this fixes the post select AJAX callback. This bug has
likely existed since the hook was introduced.

See 2e414d3b732f708e20b27354e84dab24f313a22b and
17ef5bc9efd9d90ea38ffdec2eadb7625bd682e9

See #545

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -129,6 +129,19 @@ class Shortcode_UI {
 	 * @return array
 	 */
 	public function get_shortcodes() {
+
+		if ( ! did_action( 'register_shortcode_ui' ) ) {
+
+			/**
+			 * Register shortcode UI for shortcodes.
+			 *
+			 * Can be used to register shortcode UI only when an editor is being enqueued.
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@wp-shortcake/maintainers thoughts on this approach? While it seems a bit hacky, I don't know that we have a better option, as there's a bit of a conceptual difference between editor scope and admin ajax scope — admin ajax never knows which editor the request is in context to.

mattheu:
hmm yeah - I think this looks OK to me.

goldenapples:
Ah, I see the issue here.

Fieldmanager does something nice with this where some basic context about the UI is passed into every Ajax request so the correct handlers can be loaded on the back end. That might be a bit more elegant, but for now the array of shortcodes is all registered globally, so there's no reason to go that route yet.

Issue #543, opened 2015-11-06
Radio is broken.


nextgenthemes:
``` php
            array(
                'attr'  => 'align',
                'label' => __('Alignment', $this->plugin_slug ),
                'type'  => 'select',
                'options' => array(
                    ''       => __( 'Default', $this->plugin_slug ),
                    'none'   => __( 'None', $this->plugin_slug ),
                    'left'   => __( 'Left', $this->plugin_slug ),
                    'right'  => __( 'Right', $this->plugin_slug ),
                    'center' => __( 'center', $this->plugin_slug ),
                ),
            ),
```

This is part of my register array. Works Perfectly fine, as soon as I replace type with radio. The UI shows the radio boxes as expected but no matter what I select its not added to the short-code.

jordif:
Hi,

I'm experiencing the same issue both with checkboxes and radio buttons. The cause seems to be the following commit:

https://github.com/wp-shortcake/shortcake/commit/9f5890f3d204fadb3f74286704357e0dbb3f3bd5

(Rolling back to the previous array of events fixes this)

Jordi

danielbachhuber:
Thanks for the diagnosis, @jordif. We'll get this fixed in v0.6.2. We'd recommend rolling back to v0.5.0 if this issue impacts you in production.

danielbachhuber:
@jordif Just released v0.6.2 to fix this issue. Thanks again for the report!

PR #548, opened 2015-11-12
Listen for "change" event on radio buttons and checkboxes

goldenapples:
Because these inputs don't actually have a value, they don't trigger an
"input" event. Listening for the "change" event as well on certain kinds
of inputs ensures that the model will be updated appropriately.

See #543 

Fixes an issue which was introduced in 9f5890f, in trying to listen to "input" events for better responsivity in text fields and textareas.


goldenapples:
Thanks for the diagnosis, @jordif.

danielbachhuber:
:100:

Issue #131, opened 2015-02-02
TinyMCE button / Quicktag rather than Add Media


bobbingwide:
It's probably no big thing but it's not immediately obvious that Add Media would be used to create a shortcode. It would be nice to be able to support display of the "Insert Post Content" dialog through a toolbar button.

In my oik plugin I defined a TinyMCE button in the Visual editor and a Quicktag for the Text editor. 
Both buttons are optional; the UI is not enforced. 

I imagine this is related to #115 and #94 

BTW. While investigating a "good reason" for doing this I created a custom post type that doesn't support thumbnail. I noticed two problems when shortcake was activated. 
- The Add Media button appeared but didn't do anything
- The were no quicktag buttons on the Text editor.

mattheu:
We originally had a separate button, and decided to role it all into the single button. I'm certainly open to UX feedback on this though. However I think we should be consistent - pick one and go with it.

If we stick with the current - It should definitely try to handle your situation.

bfintal:
I agree that this should be inside a TinyMCE button instead though, since strictly speaking, a post element isn't media. The "Insert Post Element" integrates great with the media manager, so it should be kept that way. When the TinyMCE button is clicked though it should open the media manager with the "Insert Post Element" pre-selected.

chrisvanpatten:
I agree and think a TinyMCE button makes more sense, or at least as its own button next to Add Media. Whether the Insert Post Element tab gets broken out into its own modal or not is beyond the scope of my opinions though :)

bfintal:
> or at least as its own button next to Add Media 

+1

ghost:
This snippet will add a button next to `Add media`, clicking the button will open the default WP Media dialogue on the `Add Post Element` tab.

```
/**
 * Add Post Element button.
 */

function alpha_add_post_element_button_markup() {
  echo '<a href="#" class="button shortcake-add-shortcode"><span class="wp-media-buttons-icon dashicons dashicons-migrate"></span> ' . __( 'Add Post Element', 'alpha' ) . '</a>';
}
add_action( 'media_buttons', 'alpha_add_post_element_button_markup', 100 );

// Action on button click

add_action( 'wp_enqueue_media', 'alpha_add_post_element_button_action' );

function alpha_add_post_element_button_action(){
  add_action( 'admin_print_footer_scripts', 'alpha_add_post_element_button_scripts', 999);
}

function alpha_add_post_element_button_scripts(){ ?>

[...]

PR #551, opened 2015-11-18
Add "Add Post Element" to media buttons

goldenapples:
Includes an "Add Post Element" to the media buttons array, which pops
open the Insert Post Element frame of the media modal.

See #131

props to @ninetienne and @Vaince for code snippets


goldenapples:
Clicking this button:

screen shot 2015-11-18 at 1 51 07 pm

will bring up the "Insert Post Element" state of the media modal.

Curious to hear feedback on this. I know there were concerns raised about adding more buttons to the editor, but on the other hand, this has been a very popular request. We have time before the next release to put some thought into the UI, so it would be good to get some community input. I think adding a media button but not an MCE button is a good middle ground between discoverability and clutter.

danielbachhuber:
> Curious to hear feedback on this.

I dig it. :+1: from me.

mattheu:
I like the button. But I'm not so sure about the UX of having 2 links to the same modal. I think we need to go further and split this out from the existing media modal completely. See https://github.com/wp-shortcake/shortcake/pull/582

khromov:
Putting this in the media dialogue has always felt a bit awkward, and we've gotten feedback from multiple editors that it's a useless extra click, so +1 this!

0aveRyan:
Small issue I've encountered, still working on a fix -- The media modal retains its' state on close. 

The script works initially to launch into the right view, but if an image or other media element is inserted first, the modal retains state and will return to image, gallery, etc. Apparently WordPress Core intends to retain state (see: https://core.trac.wordpress.org/ticket/27480). I've spent a few minutes attempting to tap into the modal and reset any existing states and force `shortcode-ui`, but haven't been successful yet.

0aveRyan:
@goldenapples right before the `wp.media.editor.open( editor, options )`, if you add this: 

```
wp.media.editor.remove( editor );
```

The modal will always trigger the right state, even if there's a stored state from Inserting an image or other view.

goldenapples:
Thanks @0aveRyan - that seems to work perfectly for triggering the correct state of the modal. I think we'll also need to find an appropriate place of resetting the modal state when the shortcode-ui modal is closed; otherwise, after once clicking "Add post element", when the user tries to add media, they'll get the "Insert Post Element" state.

0aveRyan:
@goldenapples that would require a more invasive change to the callback where Shortcake inserts the shortcode or the close method is triggered. 

I agree and personally prefer the state to reset on close / insertion as well... but the discussion on that [WP Core ticket](https://core.trac.wordpress.org/ticket/27480) indicates they prefer the state is retained, so while it definitely makes sense to unset for this button, think we should get feedback from someone on wiping out state elsewhere before exploring that.

goldenapples:
I merged in master to fix the build lint error; this is ready for a review/merge now.

mattheu:
Looks good. Tested and works well.

PR #555, opened 2015-11-20
vars being global for no good reason

kraftner:
I guess at least in the second case this is a matter of `;`instead of `,`


danielbachhuber:
Build fails for unrelated reason. Thanks for the PR!

Issue #559, opened 2015-11-30
Bug: Clicking the edit icon requires two clicks to open the Shortcake modal


goldenapples:
After the shortcake modal has been opened once, clicking the pencil icon on any post element in the TinyMCE editor requires two clicks to open the '{Shortcode} Details` frame of the media editor.

This is specific to this frame having been last open, and trying to open it again with the same or a different shortcode. I suspect that the `currentShortcode` property isn't set correctly when first trying to open that modal, or that it isn't unset on closing the `wp.media.editor` the first time, and this causes an error setting it later. 

I don't see any console errors or obvious causes for the issue.

jesperbjerke:
this is the same bug reported here #501

goldenapples:
Thanks! I knew I'd seen it brought up before, but couldn't find it by title. yes, it looks like the same bug, and your debugging there is a good start.

danielbachhuber:
:hammer:

minimaluminium:
Has this issue been fixed yet?

goldenapples:
@sodbileg Yes, this issue was fixed with #560, and should be resolved in the master branch. We haven't cut a new release since then, so the problem exists in 0.6.2, the latest released version.

oritiger:
@goldenapples Hi, I encountered the same issue. Are there any plan for the next release?

PR #560, opened 2015-12-02
Reset the shortcode regexp's index between each match

goldenapples:
Because the shortcode Regexp defined by `wp.shortcode.rexexp` is a
global matcher, each time it is `exec()`'d, it preserves the last index
searched up to, and begins searching from that index of the target
string the next time it is run over that target string.

This fixes that problem by resetting the lastIndex counter to 0 before
executing the regex.

See #559

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -162,6 +162,7 @@ var $ = (typeof window !== "undefined" ? window['jQuery'] : typeof global !== "u
 
 describe( 'Shortcode View Constructor', function(){
 
+
 	it( 'Persists inner_content when parsing a shortcode without inner_content attribute defined', function(){
 		var data = {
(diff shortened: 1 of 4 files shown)


goldenapples:
The build failure is unrelated, stemming from new expectations in WPCS PHP code sniffs.

danielbachhuber:
> The build failure is unrelated, stemming from new expectations in WPCS PHP code sniffs.

You can use this fix https://github.com/WP-API/WP-API/pull/1774

goldenapples:
> You can use this fix

Thanks, that simplifies things. Was thinking I'd have to make a whitespace-only pull request to keep the build from failing. Fixed in #561.

goldenapples:
@danielbachhuber review/merge?

PR #568, opened 2015-12-18
Debug ajax request when other plugins already add a get parameter to ajaxurl

Vaince:
Hello,

This pull request checks if ajaxurl already contains get parameters that can be added by other plugins (such as WPML).

Xoxo.
Vaince.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -365,7 +365,7 @@ var Fetcher = (function() {
 			return;
 		}
 
-		var request = $.post( ajaxurl + '?action=bulk_do_shortcode', {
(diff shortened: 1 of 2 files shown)


goldenapples:
Interesting. What do these additional query parameters do?

I'm wondering if it makes sense to build the ajax endpoint ourselves and pass it through the localization object so that it was available as something like `shortcodeUIData.endpoints.bulk_do_shortcode`, rather than relying on WP's `ajaxurl` variable.

danielbachhuber:
> ajaxurl already contains get parameters that can be added by other plugins (such as WPML).

I think this is a problem of WPML, and not something Shortcake should be expected to fix.

goldenapples:
I'd like to test including the `action` parameter in the POST body rather than building out the URL to it ourselves. That should fix the use case mentioned here - and when we update to using a REST endpoint for previews, it'll be a non-issue anyways.

goldenapples:
Closing this in favor of #653, which should address the same issue using core functionality.

PR #569, opened 2016-01-08
Cut Travis' workload in half

danielbachhuber:
Prevent duplicate builds from running on pull requests by only
triggering branch builds on master


mattheu:
Sounds reasonable

PR #580, opened 2016-02-29
delete shortcode preview

mattheu:
I'd like to clean up the codebase - in particular removing older and unused functionality.

This preview code is no longer in use. I'd like to have a generic preview handler that could be used by the MCE view, but currently this isn't it. 

I say remove this - we've always got it in git history if needed.


danielbachhuber:
:ship: :ship:

PR #581, opened 2016-02-29
Dont bundle wp scripts

mattheu:
The JS tests rely on some core WP JS. The solution so far was to bundle a copy of the scripts. 

This is obviously a pain.
They get out of date easily. 
We depend on quite a few, and may need more in future. 

I'd much rather just reference the core files so we can run the tests against whichever version of WP you have.

PROBLEM - how to determine the relative path to these files? I've used an `abspath` variable in the Gruntfile, which you can modify as needed. EG if Shortcake is bundled in a theme on WP VIP.

Is there a better solution to this problem? Should I just keep bundling the files and keep them all updated? 

Would love some thoughts on this.


mattheu:
One idea - we can use the test WP install used by PHP unit.

This does add a barrier to running the tests though, so I've removed them from the standard 'scripts' task. We will have it running on Travis so I think this is OK.

mattheu:
OK I think I've got a solution. Use grunt option to pass the custom path.

eg
`grunt jasmine --abspath="/srv/www/wp-test"` to run the tests using a test install.
`grunt jasmine --abspath="../../../"` to run the tests using the current install, if shortcake is in the plugins directory.

It will default to `/tmp/wordpress`

mattheu:
This is updated and will read the location from the environment variable if not explicitly passed.

Also note that the test for dashes in an attribute was previously unsupported by WordPress, but using the scripts from core means we've got this fix.

Anything further need doing here?

Gruntfile.js
```diff
@@ -2,7 +2,20 @@ module.exports = function( grunt ) {
...
+	var banner   = '/**\n * <%= pkg.homepage %>\n * Copyright (c) <%= grunt.template.today("yyyy") %>\n * This file is generated automatically. Do not edit.\n */\n';
+
+	// Path to WordPress install. Either absoloute or relative to this plugin.
+	// Change this by passing --abspath="new/path" as a grunt option.
+	var abspath;
+
+	if ( grunt.option( "abspath" ) ) {
+		abspath = grunt.option( "abspath" );
+	} else if ( 'WP_DEVELOP_DIR' in process.env ) {
+		abspath = process.env.WP_DEVELOP_DIR;
```
danielbachhuber: I think this needs to be `process.env.WP_DEVELOP_DIR + '/src/'`

```
$ echo $WP_DEVELOP_DIR
/srv/www/wordpress-develop.dev/
```

Gruntfile.js
```diff
@@ -2,7 +2,20 @@ module.exports = function( grunt ) {
...
+	} else {
+		abspath = '/tmp/wordpress';
```
danielbachhuber: Can we perform a check to ensure `abspath` actually exists, and show a helpful error if it doesn't?

danielbachhuber:
I get this error when I try to run the tests locally:

```
local ➜  shortcake git:(dont-bundle-wp-scripts) ✗ grunt jasmine
/srv/www/wordpress-develop.dev/src
Running "jasmine:shortcodeUI" (jasmine) task
Testing jasmine specs via PhantomJS

>> TypeError: 'undefined' is not an object (evaluating 'Backbone.Model') at
>> js-tests/build/specs.js:557
>> js-tests/build/specs.js:568
>> js-tests/build/specs.js:1 s
>> js-tests/build/specs.js:1
>> js-tests/build/specs.js:2
>> js-tests/build/specs.js:1 s
>> js-tests/build/specs.js:1 e
>> js-tests/build/specs.js:1210
Warning: No specs executed, is there a configuration error? Use --force to continue.
```

What am I doing wrong?

mattheu:
I've updated this to include a task that will perform the check, and throw a fatal if the path isn't found.

I've also added a `grunt tests` command that will run this check first.

Finally I've updated the docs to indicate how to run these tests.

mattheu:
@danielbachhuber I'd like to get this one live - so am interested in checking whether you are still having problems.

Regarding your error, it looks like WP core files are not found. If you run `grunt test` instead of `grunt jasmine` there is an additional check to ensure the location is correct. I'm don't think its possible to hook this in on the standard `grunt jasmine` command. We will document this.

The current logic is as follows
- pass path like so: grunt test --abspath="/path/to/wp/install"`
- Reads environment variable WP_CORE_DIR
- Reads environment variable WP_DEVELOP_DIR and looks in `WP_DEVELOP_DIR  + '/src'`
- Defaults to `/tmp/wordpress`

Let me know if you think this needs to be improved.

danielbachhuber:
> Let me know if you think this needs to be improved.

Given the additional complexity involved in getting the test suite to run, I still think it's better to just copy the core files into the test suite.

mattheu:
OK. The big problem with this is keeping them up to date with the latest code, as there's quite a few and we may need even more in future. How about a grunt task to automate updating them instead? We can just commit these changes.

danielbachhuber:
> How about a grunt task to automate updating them instead? We can just commit these changes.

💯 

I think this is a much better solution to the general problem. This way, it's just us that have to deal with the burden of the dependency, not everyone that wants to contribute / write tests.

mattheu:
Updated the PR to add this task and include all the necessary scripts.

.travis.yml
```diff
@@ -22,7 +22,7 @@ cache:
 
 before_script:
   - |
-    if [[ "$WP_TRAVISCI" == "travis:phpunit" ]] ; then
+    if [[ "$WP_TRAVISCI" == "travis:phpunit" ]] || [[ "$WP_TRAVISCI" == "travis:js-tests" ]]; then
```
danielbachhuber: Does this still need to run for `js-tests`?

Gruntfile.js
```diff
@@ -186,6 +187,43 @@ module.exports = function( grunt ) {
...
+	 */
+	grunt.registerTask( 'updateJasmineCoreScripts', function() {
+
+		var abspath = grunt.option( "abspath" );
+
+		if ( ! grunt.file.exists( abspath + '/wp-includes' ) ) {
+			grunt.fail.fatal( 'WordPress test install not found. See readme for more information. Currently looking here: ' + abspath );
+		}
+
+		var scripts = [
```
danielbachhuber: Could we define this variable further up, and use it in our vendor definition too?

Gruntfile.js
```diff
@@ -196,9 +234,10 @@ module.exports = function( grunt ) {
 	grunt.loadNpmTasks( 'grunt-contrib-jasmine' );
 	grunt.loadNpmTasks( 'grunt-contrib-jshint' );
 
-	grunt.registerTask( 'scripts', [ 'browserify', 'jasmine', 'jshint' ] );
+	grunt.registerTask( 'scripts', [ 'browserify', 'jshint' ] );
 	grunt.registerTask( 'styles', [ 'sass', 'postcss' ] );
 	grunt.registerTask( 'default', [ 'scripts', 'styles' ] );
+	grunt.registerTask( 'test', [ 'jasmine' ] );
```
danielbachhuber: Did you mean to keep this split out as a separate task? I don't think there's harm in just running it as a part of the `scripts` task.

README.md
```diff
@@ -31,6 +31,13 @@ New in 0.4.0 is the ability to [attach javascript functions to event attribute u
 
 Shortcake doesn't support custom key=>value pairs as shortcode attributes because it isn't a great user experience.
 
+## Running tests ##
+
+We have test coverage for PHP using PHPunit, and JavaScript using Jasmine.
+
+*Running tests locally*
+
+Jasmine tests can be run using `grunt test`. Note that the JavaScript tests need access to a WordPress installation as the code depends on some core JS files. It will pick up your WP_DEVELOP_DIR environment variable as used by PHPunit, or you can pass use the `abspath` option to pass the manually like so `grunt test --abspath=/path/to/wp-install`. Note - it only needs access to the files, and not a fully functional site.
```
danielbachhuber: I think this isn't quite accurate.

danielbachhuber:
@mattheu Noted a few things

mattheu:
Updated to fix issues - I knew I should have just closed this and opened a new PR ;)

danielbachhuber:
> Updated to fix issues - I knew I should have just closed this and opened a new PR ;)

:)

PR #582, opened 2016-02-29
Rewrite - Standalone frame

mattheu:
Following some of the feedback we've had, I've pulled the 'Add post element' menu item out of the 'add media' modal, and into its own frame. 

This is a bit of a big change, but would be interested to get some feedback on it.

From a UX perspective, I like it. Code wise, I like this too. Feels quite neat. 

Heres a rundown of the changes.
- Backwards compatibility - should be good, unless you're doing a lot of custom stuff.
- Consolidate `views/shortcode-ui.js` and `views/media-frame.js` into a single view. These were too tightly coupled and couldn't really be used independently of one another. 
- The code for the new frame very closely follows the code for the core `MediaFrame`.
- The new frame can now be used anywhere, and is not really coupled with TinyMCE. I've allowed you to pass a custom `insertCallback` to do whatever you need. You would of course need to do a bunch of stuff yourself such as preview, edit, buttons etc. But theoretically its possible. 
- Cleanup of code. Tried to trim it all down, and reuse some core stuff as much as possible. 

Despite the complete rewrite of the media-frame code, it should still function very much as before.

NOTE - this DIFF is massive. But i've got a few other pull requests that can go in before this and will make it much more readable.


Gruntfile.js
```diff
@@ -2,7 +2,12 @@ module.exports = function( grunt ) {
 
 	'use strict';
 	var remapify = require('remapify');
-	var banner = '/**\n * <%= pkg.homepage %>\n * Copyright (c) <%= grunt.template.today("yyyy") %>\n * This file is generated automatically. Do not edit.\n */\n';
+	var banner   = '/**\n * <%= pkg.homepage %>\n * Copyright (c) <%= grunt.template.today("yyyy") %>\n * This file is generated automatically. Do not edit.\n */\n';
+
+	// Path to WordPress install. Either absoloute or relative to this plugin.
+	// Change this by passing --abspath="new/path" as a grunt option.
+	var abspath = grunt.option( "abspath" ) ? grunt.option( "abspath" ) : '/tmp/wordpress';
```
danielbachhuber: Would it be possible to sniff the `WP_DEVELOP_DIR` environment variable too, so it works interchangeably with how PHPUnit tests work?

css/shortcode-ui.css
```diff
@@ -20,16 +30,16 @@
...
       height: 120px;
       font-size: 64px; }
-      .add-shortcode-list .shortcode-list-item .add-shortcode-list-item-icon .dashicons, .add-shortcode-list .shortcode-list-item .add-shortcode-list-item-icon img {
+      .insert-shortcode-list .add-shortcode-list .shortcode-list-item .add-shortcode-list-item-icon .dashicons,
+      .insert-shortcode-list .add-shortcode-list .shortcode-list-item .add-shortcode-list-item-icon img {
         position: absolute;
         top: 50%;
         left: 50%;
         -webkit-transform: translate(-50%, -50%);
-            -ms-transform: translate(-50%, -50%);
```
danielbachhuber, line 32: Should this vendor prefix be removed?

js/build/shortcode-ui.js
```diff
@@ -1685,372 +1547,253 @@ var wp = (typeof window !== "undefined" ? window['wp'] : typeof global !== "unde
...
-			this.refreshShortcodes( sui.shortcodes );
-		} else {
-			this.refreshShortcodes( this.controller.search( event.target.value ) );
+		_.bindAll( this, 'select', 'reset' );
+
+		this.options = _.defaults( this.options, {
+			state:          'shortcode-ui',
+			modal:          true,
+			title:          'Insert',
+			updateTitle:    'Update',
```
danielbachhuber: We should make sure these strings are localizable out of the box.

js/src/views/insert-shortcode-list.js
```diff
@@ -6,37 +7,84 @@ var insertShortcodeListItem = require('sui-views/insert-shortcode-list-item');
...
-	}
+		}.bind(this) );
+
+	},
+
+	search: function( s ) {
+
+		if ( s && s.length ) {
+
+			var pattern = s.replace( /[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&" );
```
danielbachhuber, line 72: Where is this regex from?

danielbachhuber:
Lot of code changing here :) I'll check it out locally sometime in the next week or so to give it more of a click-through.

For common reference, can you share some screenshots or GIFs of how it looks?

Gruntfile.js
mattheu: I've added this too. Prioritiy is 
1. If passing abspath use that. (Allows running tests without setting up full test install)
2. If environment var defined use that
3. Else use `/tmp/wordpress`

mattheu:
Yes this is a big one :)

He'res a little video: https://www.youtube.com/watch?v=QOAlpes9rz8

Note - it includes the code from this which can go in first and is also fairly large: https://github.com/wp-shortcake/shortcake/pull/581/

js/src/views/insert-shortcode-list.js
mattheu, line 72: Problem is that currently we just pass the whole search query when creating to the regex. Try seaching for `\` or `*` and you get a JS error. 

This is trying to escape anything in the search query before using to create the Regex.

js/build/shortcode-ui.js
mattheu: Fixed. These should be empty strings as defaults, with localized labels passed in when creating the view. I've updated this to ensure they are.

css/shortcode-ui.css
mattheu, line 32: I guess autoprefixer has decided it's no longer necessary? I think it's fine to go with that - if we need to be more explicit about which browsers we're targeting - we should update the autoprefixer task - but I'd tackle that separately.

PR #583, opened 2016-03-02
Use WP Rest API infrastructure for previews

mattheu:
How about using the rest API infrastructure in core for previews? 
- I don't particularly think it has any real advantages? The way it handles sanitization of args is nice though.
- doing a POST request to `get preview` does feel a bit odd. Any issues using GET instead?

Also deleted the legacy 'fetch' code which is no longer used - including tests. Becomes apparent that we don't have tests for the fetcher which we probably should - but this can be handled separately.


inc/class-shortcode-ui.php
```diff
@@ -244,9 +269,13 @@ public function enqueue() {
 				'insert_content_label'              => __( 'Insert Content', 'shortcode-ui' ),
 			),
 			'nonces'     => array(
-				'preview'        => wp_create_nonce( 'shortcode-ui-preview' ),
+				'wp_rest'        => wp_create_nonce( 'wp_rest' ),
 				'thumbnailImage' => wp_create_nonce( 'shortcode-ui-get-thumbnail-image' ),
 			),
+			'urls'       => array(
+				'preview'     => home_url( '/wp-json/shortcode-ui/v1/preview' ),
```
danielbachhuber: We should use `rest_url()` here, to accommodate non-pretty permalinks as well.

inc/shortcode-ui.php
```diff
@@ -0,0 +1,421 @@
+<?php
+
+$_GLOBALS['shortcode_ui_shortcodes'] = array();
```
danielbachhuber: I think it's `$GLOBALS` — and why do we need to globalize this here?

mattheu: Ah damn this shouldn't be in this PR :)

danielbachhuber:
Looks pretty good!

> I don't particularly think it has any real advantages?

Modernization is the key benefit, I think. Also, when batch `GET` is added to core, we'll be able to make use of it for batch previews.

> doing a POST request to get preview does feel a bit odd. Any issues using GET instead?

Given we aren't performing a write operation, I think we should be using `GET`.

inc/shortcode-ui.php
```diff
@@ -0,0 +1,421 @@
...
+
+add_action( 'admin_enqueue_scripts',     'shortcode_ui_action_admin_enqueue_scripts' );
+add_action( 'wp_enqueue_editor',         'shortcode_ui_action_wp_enqueue_editor' );
+add_action( 'media_buttons',             'shortcode_ui_action_media_buttons' );
+add_action( 'wp_ajax_bulk_do_shortcode', 'shortcode_ui_handle_ajax_bulk_do_shortcode' );
+add_filter( 'wp_editor_settings',        'shortcode_ui_filter_wp_editor_settings', 10, 2 );
```
danielbachhuber: Are we removing the singleton then?

mattheu: We discussed this somewhere... but this is for a separate PR ;)

mattheu:
Removed some stuff I committed by accident. 

Updated PR to use `rest_url`

mattheu:
> batch GET

Yeah that would be really neat

inc/class-shortcode-ui.php
```diff
@@ -68,9 +68,34 @@ private function setup_actions() {
 		add_action( 'admin_enqueue_scripts',     array( $this, 'action_admin_enqueue_scripts' ) );
 		add_action( 'wp_enqueue_editor',         array( $this, 'action_wp_enqueue_editor' ) );
 		add_action( 'wp_ajax_bulk_do_shortcode', array( $this, 'handle_ajax_bulk_do_shortcode' ) );
```
danielbachhuber: We can remove this registration, right?

danielbachhuber:
@mattheu Can you fix the failed build? Looks good to me otherwise.

@goldenapples ?

inc/class-shortcode-ui.php
mattheu: Oh yep

inc/class-shortcode-ui.php
```diff
@@ -366,38 +394,63 @@ private function render_shortcode_for_preview( $shortcode, $post_id = null ) {
...
-	 * Get a bunch of shortcodes to render in MCE preview.
+	 * Get a preview for a single shortcode to render in MCE preview.
 	 */
-	public function handle_ajax_bulk_do_shortcode() {
-
-		if ( is_array( $_POST['queries'] ) ) {
-
-			$responses = array();
+	public function handle_shortcode_preview( WP_REST_Request $request ) {
+		return $this->get_shortcode_preview( $request->get_param( 'query' ) );
```
goldenapples: I'm not sure what this function is doing, but we don't currently have a `get_shortcode_preview()` method here...

I think this functionality might be vestigial anyways, currently all previews are bulk requests even if they're only for one shortcode.

inc/class-shortcode-ui.php
```diff
@@ -244,9 +268,13 @@ public function enqueue() {
 				'insert_content_label'              => __( 'Insert Content', 'shortcode-ui' ),
 			),
 			'nonces'     => array(
-				'preview'        => wp_create_nonce( 'shortcode-ui-preview' ),
+				'wp_rest'        => wp_create_nonce( 'wp_rest' ),
```
goldenapples, line 250: Unless I'm missing something, we're not checking nonces on any of the REST request handlers. Should we be?

I think there might be a benefit to making this unauthenticated, as then it would be easier to reuse the functionality in external editors and the like. However, that requires a much more solid approach to security, as it opens an attack surface for anyone to compel a server to render an arbitrary shortcode. Given that people still use things like `[php]` shortcodes, this gets sketchy pretty fast.

goldenapples:
This looks much cleaner than before. I like all the changes here, and doing this through the REST API makes it potentially discoverable for other uses.

inc/class-shortcode-ui.php
mattheu, line 250: The nonce check is done by the rest API itself. Otherwise the request is treated as unauthenticated - which we are checking for when we do the edit_post caps check. 

Interesting thoughts RE opening this up for others. However I think we should keep it authenticated for now. There would be quite a big potential for exploiting this. 

@danielbachhuber can you flag and endpoint as requiring auth? Perhaps in the schema?

danielbachhuber, line 250: Yes, we need to make sure the endpoint implements authentication. We'll want to include a `permissions_callback` for each endpoint — and it would be good to write tests around this as well.

mattheu:
Added a permission callback to the requests to check the user has the cap to edit the post.

mattheu:
TODO 
- add some tests.

mattheu:
I've added some tests for this. Should be good to go. @danielbachhuber would be good if you could give this a once over.

westonruter:
I have a concern about using the REST API for this. Namely, shortcodes may render differently depending on the URL that the shortcode appears on. In the Customizer, the Selective Refresh API makes requests to render partials by making an Ajax request to the actual URL that the partial (e.g. widget) appears on, and a `template_redirect` intercepts the request to just render the requested partials as opposed to loading the normal template. This ensures that when the partials render, they render in the expected context.

mattheu:
Hmm thats a good point - and sounds like a good workaround. This would affect the current implementation too.

mattheu:
Lets push this to a future release. Too big for 0.7.0

PR #587, opened 2016-03-23
Finnish translation

TeemuSuoranta:
Here's the Finnish translation of plugin. There was no way to translate it through WordPress.org so I did it here. As a side note I noticed that the link to Author URI is broken (http://next.fusion.net/tag/shortcode-ui/). I left it as is in the translation, though.


mattheu:
Thanks for this. Sorry not merged sooner. Failing CI was an issue on master, which is now fixed so I'm going to merge.

PR #589, opened 2016-03-31
Fix search matching index

### LINKED ISSUE #458: Insert Post Element Search filter sometimes omits items
The search facility misses out some matches.

I quite like the search box on the Insert post element dialog. 
But I'm not sure how it find matches. 
I've noticed that sometimes it fails to display what I consider to be matches. 

e.g. I have a number of shortcodes starting `bw_f`
If I type `bw_field` then I'd expect to see both `bw_field` and `bw_fields`
but `bw_fields` doesn't show until I type the `s`

If I just type `f` then the `bw_follow_me` shortcode does not appear in the list.

BTW: I've not looked at the code.

khromov:
Fixes #458

Shortcake elements whose name start with the same string might not get included in search due to the RegExp being reused. 

Example: Two content elements:
- Character List
- Character Grid

Search for: Character - only one of the items shows up.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -63,6 +63,7 @@ var MediaController = wp.media.controller.State.extend({
 	search: function( searchTerm ) {
 		var pattern = new RegExp( searchTerm, "gi" );
 		var filteredModels = sui.shortcodes.filter( function( model ) {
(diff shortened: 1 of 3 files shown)


mattheu:
Thanks - I've run into this before. This is a neat fix.

PR #591, opened 2016-04-04
Use array of options for fields

mattheu:
Bug - my select field options are all shown in a random order. I have a large-ish list and this is pretty important.

Options for select fields are currently passed as an object like this `{ value: 'Label' }`.

This isn't great because objects don't really have an order. The solution is that these should really be an array like so `[ { value: 'value', label: 'Label' } ]`

I've updated the template to handle this format and added a backwards compatability shim toconvert the old object format into an array.


mattheu:
Any thoughts on this?

mattheu:
Added some tests for this bug.

mattheu:
Ready for review/merge

goldenapples (commented):
This looks good, and I like the compatibility shim in there. Merging this, so that we can unblock a few other issues. We should make sure to update documentation as needed after this.

PR #592, opened 2016-04-04
fix failing CS

mattheu:
CS is failing due to a badly formatted concat.


mattheu:
Actually fixed here: https://github.com/wp-shortcake/shortcake/pull/589

mattheu:
Fixed by https://github.com/wp-shortcake/shortcake/pull/589

PR #596, opened 2016-04-20
Update pot file

hideokamoto:
Hello.

I think pot file is old.
So, I update it.

thank.


goldenapples:
Confirmed that this .pot matches the current state of the plugin (adds several new strings which weren't included in the last .pot file, which was generated before v0.6.0).

Issue #415, opened 2015-08-09
New Field: Term Select


dev4press:
This field would be similar to post_select, but for terms selection.. User register the field and set the taxonomy terms can be choosen from.

danielbachhuber:
Seems reasonable — as well as User Select

jurajk:
would love to see both these fields in future releases.

johnnypea:
+1

marcusforsberg:
+1 :)

misfist:
Taxonomy select would be super useful. +1

mattheu:
We've got this now!

misfist:
I have this code and the select isn't rendering.

``` php
public function image_gallery_shortcode_ui() {
    $fields = array(
        array(
            'label'    => esc_html__( 'Select Category', 'cvar-core' ),
            'attr'     => 'slug',
            'type'     => 'term_select',
            'query'    => array( 'taxonomy' => 'media_category' ),
            'multiple' => true,
        ),
    );

    /*
     * Define the Shortcode UI arguments.
     */
    $shortcode_ui_args = array(
        'label'         => esc_html__( 'Add Image Gallery', 'cvar-core' ),
        'listItemImage' => 'dashicons-format-gallery',
        'post_type'     => array( 'post', 'page' ),
        'attrs'         => $fields,
    );

    shortcode_ui_register_for_shortcode( 'image-gallery', $shortcode_ui_args );
}
```

Is there an example of this field's usage?

BronsonQuick:
Hey @misfist trying altering:

```
'query'    => array( 'taxonomy' => 'media_category' ),
```

to:

```
 'taxonomy' => array( 'media_category' ) 
```

There's an example of this fields usage in: https://github.com/humanmade/shortcake/blob/52882de66decb6a6d5d4d4b59451dcf2f59ff3e1/dev.php#L189

I hope that helps you! 😄

BronsonQuick:
Or if you prefer you can just do:

```
 'taxonomy' => 'media_category' 
```

misfist:
@BronsonQuick - Yes, this totally helps! Though, I think the main problem was that I was using the [WP repo version](https://wordpress.org/plugins/shortcode-ui/) (wpackagist.org, actually) of Shortcake UI with which it didn't work. 🔢 .

I updated my composer file to source from this repo.

Thanks!

misfist:
Is there any way to return the term slug rather than the term id?

[...]

PR #604, opened 2016-05-19
Add Term Select

BronsonQuick:
Adds a `term_select` field for #415.


inc/fields/class-field-term-select.php
```diff
@@ -0,0 +1,184 @@
...
+		// Check to see if the taxonomy has been set.
+		foreach ( $shortcode['attrs'] as $attr ) {
+			if ( $attr['attr'] === $requested_attr && isset( $attr['taxonomy'] ) ) {
+				$taxonomy_type = $attr['taxonomy'];
+			}
+		}
+
+		// If the user hasn't specified a taxonomy then set it to null.
+		if ( empty( $taxonomy_type ) ) {
+			$taxonomy_type = null;
```
mattheu, line 140: I wonder if this should just do `wp_send_json_error` at this point if no taxonomy is found. As far as I can see, you need the 'taxonomy' to work properly anyway.

inc/fields/class-field-term-select.php
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		if ( $taxonomy_type && ! current_user_can( $taxonomy_type->cap->assign_terms ) ) {
```
mattheu: Could this cause us to get into a situation where a user is able to insert the shortcode using the UI, but not actually add anything because they fail this check.

Maybe we need to check this before adding the shortcode UI at all? Or maybe this isn't necessary as we're not really assigning the terms.

inc/fields/class-field-term-select.php
BronsonQuick, line 140: I did that so that the `null` flows through eventually to get_terms which then means the default `post_tag` is used if a tax isn't added by the developer that way it's it's a smart default.

inc/fields/class-field-term-select.php
BronsonQuick: Good catch. I'll change this now!

BronsonQuick:
@mattheu Thanks for the review. I updated two more things but I didn't do https://github.com/wp-shortcake/shortcake/pull/604/files#r64215458 I'll see what you think of my comment and if that doesn't make sense then I'll remove that and add in your suggestion. Taa!

inc/fields/class-field-term-select.php
mattheu, line 140: In that case, I'd rather be explicit about the defaults. Also I don't think `null` is permitted by `wp_count_terms`.

BronsonQuick, line 140: Just tested this and null is fine

BronsonQuick:
@mattheu I've just updated this to pass the taxonomy and I've made it so that if not search term is passed it'll return the first 10 results.

mattheu:
Thanks for your work on this! Looks good to me.

Issue #242, opened 2015-03-26
Allow for HTML in the 'description' attribute when registering shortcode


cjmaio:
It would be incredibly useful to allow for HTML in the 'description' attribute when registering a shortcode.

For example, when I add a field called 'icon' for an end user to add in a Font Awesome icon, I want to give them the URL to the list of icons. When I tried it, I ended up with this:

```
Choose an icon for the circle beside the button. Simply copy the name from any of the <a href="http://fortawesome.github.io/Font-Awesome/icons/" alt="Font Awesome">Font Awesome icons</a> available.
```

I am not sure if that was done on purpose or not. If not, I may submit a PR with this functionality if desired.

mattheu:
Thanks for reporting. I think this is a fairly common requirement. There is a PR already - https://github.com/fusioneng/Shortcake/pull/189

When data is passed to an Underscores templates - you need to explicitly allow unescaped output. We'll look into a workaround.

danielbachhuber:
Need to get a verdict from VIP on how they'd like to handle escaping HTML in attributes.

danielbachhuber:
VIP wants to make sure all output is appropriately sanitized / escaped. In order to do this "properly", we'd need to switch away from Underscores to a templating engine that supports custom escaping callbacks.

danielbachhuber:
Actually, we can just go the easier route:

![image](https://cloud.githubusercontent.com/assets/36432/10313551/0ff9c9e8-6c05-11e5-927b-84e7d9969d4a.png)

PR #608, opened 2016-05-24
Fix coding standards. Remove redundant escaping - late escaping on ou…

dashaluna:
…tput

### DIFF
--- a/dev.php
+++ b/dev.php
@@ -116,7 +116,7 @@ function shortcode_ui_dev_minimal_example() {
 	shortcode_ui_register_for_shortcode(
 		'shortcake-no-attributes', // Shortcode tag this UI is for.
 		array(                     // Shortcode UI args.
-		                           'label' => esc_html__( 'Shortcake With No Attributes', 'shortcode-ui-example' ),
+			'label' => __( 'Shortcake With No Attributes', 'shortcode-ui-example' ),
 		)
 	);
 }
@@ -156,7 +156,7 @@ function shortcode_ui_dev_advanced_example() {
 	 */
 	$fields = array(
 		array(
-			'label'       => esc_html__( 'Attachment', 'shortcode-ui-example' ),
+			'label'       => __( 'Attachment', 'shortcode-ui-example' ),
 			'attr'        => 'attachment',
 			'type'        => 'attachment',
 			/*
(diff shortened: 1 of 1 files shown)


dashaluna:
@mattheu if you have time, have a look at this one :)

mattheu:
Thanks Dasha. I agree that escaping should be done as late as possible, but this extra escaping is not really doing any harm. 

It was added here https://github.com/wp-shortcake/shortcake/commit/586c55fad16a14769fc1bcf22e81ea1044379fc7, and the commit doesn't give much info and I don't really remember the background - but I have a reccolection that this may have been requested by VIP. If that's the case, I think its best kept in place.

dashaluna:
@mattheu How can we check with VIP? Late escaping is in accordance to VIP standards. It would be great to promote that as much as possible. Even if there is no real harm with things as they are at the moment, that shouldn't be the reason stopping us from improving it :) Given this is example usage, people are likely to copy it over and cary on with that escaping.

dashaluna:
@mattheu I just realised that we might  be talking about different things. I'm talking about coding standards and html escaping. You mentioned `encoding` in the link in previous comment :) That's different.

This PR is just cleaning up the Coding Standards.

mattheu:
Oh I think I liked to the wrong commit 😀

mattheu:
I'm going to take a look into this - I do wonder if there was some reason behind this early escaping.

dashaluna:
@mattheu thank you! Keep me posted.

dashaluna:
@mattheu any news? ;)

mattheu:
The only thing I can think of is that we are NOT escaping some things on output. Specifically, the label and descriptions in our JS templates. This is done in order to allow HTML such as `<b> and <a href="#">` in the descriptions.

Now we are manually sanitizing these strings using wp_kses_post here: https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L131 but perhaps this is an extra safety measure.

They're not doing any harm I don't think. Interested if anyone else has any thoughts.

goldenapples:
There was some discussion with VIP about this - see #242 - and they were OK with not escaping these fields on output. Since they're not not being late-escaped, we wanted to escape these fields on marking for i18n - otherwise translators could add HTML tags in translation which might break expected output.

mattheu:
I'm going to close this pull request. The small whitespace fixes were hit as part of https://github.com/wp-shortcake/shortcake/pull/613

dashaluna:
Sad face :( The point of the standards then if VIP selectively approves some exceptions...

PR #613, opened 2016-06-01
Update dev examples

mattheu:
We have a bunch of test fields in the dev example that aren't output by the shortcode. Should probably add them for completeness.

Also renamed 'background-color' to 'color' as its shorter.

Ready for merge.


goldenapples:
Adds several more translatable strings; which is a shame since we just merged #596. Otherwise it looks good.

mattheu:
Yeah thats a shame. At least it only affects the dev plugin.

PR #614, opened 2016-06-02
Add user_select field

dashaluna:


mattheu:
Happy to merge. These version numbers have gotten in a muddle, but they'll get sorted out in the next proper release.

inc/fields/class-field-user-select.php
```diff
@@ -0,0 +1,174 @@
...
+	 * Requires that the field has been correctly registered and can be found in $this->post_fields
+	 * Supports passing page number and search query string.
+	 *
+	 * @return null
+	 */
+	public function action_wp_ajax_shortcode_ui_user_field() {
+
+		$nonce               = isset( $_GET['nonce'] ) ? sanitize_text_field( $_GET['nonce'] ) : null;
+		$requested_shortcode = isset( $_GET['shortcode'] ) ? sanitize_text_field( $_GET['shortcode'] ) : null;
+		$requested_attr      = isset( $_GET['attr'] ) ? sanitize_text_field( $_GET['attr'] ) : null;
```
westonruter: Each of the `$_GET` references should have `wp_unslash()` before any sanitization is done.

inc/fields/class-field-user-select.php
```diff
@@ -0,0 +1,174 @@
...
+				$query_args = $attr['query'];
+			}
+		}
+
+		if ( isset( $_GET['page'] ) ) {
+			$query_args['paged'] = sanitize_text_field( $_GET['page'] );
+		}
+
+		if ( ! empty( $_GET['s'] ) ) {
+			$query_args['search'] = '*' . sanitize_text_field( $_GET['s'] ) . '*';
```
westonruter: `$_GET['s']` needs `wp_unslash()`

westonruter: I would suggest moving all of the accessing of the input vars to the top of this method and to store the sanitized values in regular variables.

So at the top:

``` php
$s = null;
if ( isset( $_GET['s'] ) ) {
    $s = sanitize_text_field( wp_unslash( $_GET['s'] ) );
}
```

And then down here:

``` php
if ( ! empty( $_GET['s'] ) ) {
    $query_args['search'] = '*' . $s . '*';
}
```

I think it is a best practice to sanitize early and escape late.

inc/fields/class-field-user-select.php
```diff
@@ -0,0 +1,174 @@
...
+
+	/**
+	 * Add Select2 for our UI.
+	 */
+	public function action_enqueue_shortcode_ui() {
+
+		$plugin_dir = dirname( dirname( __FILE__ ) );
+
+		wp_enqueue_script( 'select2', plugins_url( 'lib/select2/select2.min.js', $plugin_dir ), array( 'jquery', 'jquery-ui-sortable' ), '3.5.2' );
+		wp_enqueue_style( 'select2', plugins_url( 'lib/select2/select2.css', $plugin_dir ), null, '3.5.2' );
```
westonruter: Should `select2` be registered (e.g. via `wp_register_script()`) once and then just the handle enqueued here?

Here's how I've been doing it for a Customizer control that uses Select2:
https://github.com/xwp/wp-customize-object-selector/blob/423e8f62a326ba05f164b0077ce944ad2ed91bfc/php/class-plugin.php#L70-L76
https://github.com/xwp/wp-customize-object-selector/blob/423e8f62a326ba05f164b0077ce944ad2ed91bfc/php/class-plugin.php#L106-L112

inc/fields/class-field-user-select.php
dashaluna: @westonruter could you explain a bit more why we need `wp_unslash()`. Still learning :) Thank you!

inc/fields/class-field-user-select.php
dashaluna: @westonruter in the second code chunk you mean `if ( ! empty( $s ) ) {` as it can be non-empty before sanitisation and empty after?

westonruter: You're right, sorry. It could be just:

``` php
if ( $s ) {
    $query_args['search'] = '*' . $s . '*';
}
```

inc/fields/class-field-user-select.php
westonruter: @dashaluna sure, WordPress has an unfortunately legacy back-compat feature where it will force all of the input vars like `$_GET` and `$_POST` to be “slashed”. So for example, if you put in a search for “I'm happy”, then if you look at `$_GET['s']` it will actually be `I\'m happy`. 

For more information, see https://core.trac.wordpress.org/ticket/18322

inc/fields/class-field-user-select.php
dashaluna: Np, just making sure I understand it correctly :)

dashaluna:
@mattheu @westonruter alright, this should do it :) Over to you guys

mattheu:
Looks great @dashaluna Thanks!

Issue #617, opened 2016-06-07
Attachment field hook


marcin-lawrowski:
I have noticed that the hook is not triggered on attachment field when user deletes the attachment.

goldenapples:
Yup, that's definitely a bug. 

It's happening because there's a check in the `updateValue` method (which is where the hooks are triggered on other events) which returns early if there's no attachment set. We'll have to manually call `triggerCallbacks` in the `_removeAttachment` method.

goldenapples:
Closing, resolved by #619.

PR #619, opened 2016-06-07
Trigger event callbacks when removing an attachment

goldenapples:
Ensure that hooks registered on attachment fields are triggered when removing an attachment in the same way that they are when selecting or updating it.

Minor fix, connected to #617.


mattheu:
👍 Thanks!

PR #623, opened 2016-06-27
Only render description element if description

mattheu:
If no description is present, an empty element is still output, which also has a margin-bottom. 

The templates should only render them if the description actually has length.


goldenapples (approved):
This is a reasonable fix, and I can confirm no description elements are output when the description field isn't set.

PR #628, opened 2016-07-27
Extending the events base with: render_new|edit|destroy

mehigh:
I've made a quick screencast showing what was possible to be achieved with these events being present:
https://shrtm.nu/iB4k
I've tried to make use of the existing wp.shortcake.hooks.addAction using hardcoded shortcodes & attributes needing the rich text editing capabilities, but I encountered a brick wall by not being able to hook at the right moment to discard the rich text editor and update the textarea with the sanitised data (taking care of " / [ / ]) without the need to turn on encoding in the shortcode attribute. 

Usage example:

``` javascript
wp.shortcake.hooks.addAction( 'shortcode-ui.render_edit', function() { 
    console.log( 'The shortcode-ui edit screen was rendered.' );
} );
wp.shortcake.hooks.addAction( 'shortcode-ui.render_new', function() { 
    console.log( 'The shortcode-ui new screen was rendered.' );
} );
wp.shortcake.hooks.addAction( 'shortcode-ui.render_destroy', function() {
    console.log( 'The shortcode-ui screen is going to be destroyed.' );
} );
```

This handles the request here: https://github.com/wp-shortcake/shortcake/issues/616
And also inserts the needed pieces into the plug-in, in order to allow developers to make use of rich text input, as it was requested in: 
https://github.com/wp-shortcake/shortcake/issues/236

If this finds its way into core, I can publish the developments as a separate plug-in which would extend on this for others to be able to easily benefit of the rich text editing capabilities on textarea input fields.

### DIFF
--- a/js/src/utils/shortcode-view-constructor.js
+++ b/js/src/utils/shortcode-view-constructor.js
@@ -178,9 +178,13 @@ var shortcodeViewConstructor = {
 			 * Action run after an edit shortcode overlay is rendered.
 			 *
 			 * Called as `shortcode-ui.render_edit`.
+			 *
+			 * @param shortcode (object)
+			 *           Reference to the shortcode model used in this overlay.
 			 */
 			var hookName = 'shortcode-ui.render_edit';
-			wp.shortcake.hooks.doAction( hookName );
+			var shortcode = this.shortcodeModel;
+			wp.shortcake.hooks.doAction( hookName, shortcode );
 
 		}
 
--- a/js/src/utils/shortcode-view-constructor.js
+++ b/js/src/utils/shortcode-view-constructor.js
@@ -179,12 +179,12 @@ var shortcodeViewConstructor = {
 			 *
 			 * Called as `shortcode-ui.render_edit`.
 			 *
-			 * @param shortcode (object)
+			 * @param shortcodeModel (object)
 			 *           Reference to the shortcode model used in this overlay.
 			 */
 			var hookName = 'shortcode-ui.render_edit';
-			var shortcode = this.shortcodeModel;
-			wp.shortcake.hooks.doAction( hookName, shortcode );
+			var shortcodeModel = this.shortcodeModel;
+			wp.shortcake.hooks.doAction( hookName, shortcodeModel );
 
 		}
 
--- a/js/src/utils/shortcode-view-constructor.js
+++ b/js/src/utils/shortcode-view-constructor.js
@@ -173,6 +173,15 @@ var shortcodeViewConstructor = {
 
 			wp_media_frame.open();
 
+			/* Trigger render_edit */
+			/*
+			 * Action run after an edit shortcode overlay is rendered.
+			 *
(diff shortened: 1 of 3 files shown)


mehigh:
The fail check comes from a PHP dependency Travis couldn't handle. My updates only touched JS code, sorry I can't get the PHP dependencies handled.

goldenapples:
These seem like reasonable places to put hooks for a use case like this.

I think it'd be nice to have some kind of context passed in to the hooks. It looks like each of these hooks could get the current shortcode, and the view models for the attribute fields in the current shortcode model, as context. That way, you wouldn't need to be calling the the richtext.load decorators on every shortcode; you could target those hooks only to UIs that have the fields you wanted to target.

 Other than that suggestion, I think this would be good to have in the plugin.

mehigh:
Thank you, goldenapples for the feedback.

I only added the minimum required - i.e. the event itself, but since you think it's useful to pass along the shortcode and the view models, I'll make that happen and update the PR.

mehigh:
I've played a bit with this, and since we're not in a Views context, I wasn't able to get a reference to the _views in order to pass it along with the event. 
I am able to pass the shortcode model on all 3 new events.

I'm updating the PR next with these improvements. 

I had tried to place the new / edit hooks in other locations (in a views context for example), but wasn't able to reach the post layout display moment inside a View scope on the shortcode-ui.js. I was always left out with the event being passed at a too early state for it to be useful.

mehigh:
@goldenapples 
Is there anything else I need to do in order to get this merged in?
Assuming only passing the model to the events is sufficient, should I make any amends to the readme / or wiki to explain the new potential usage of the events?

Once more the Travis failure is on some PHP dependencies he didn't have available to run the php linting on. I only touched JS code which passed the linting tests.

goldenapples:
> Is there anything else I need to do in order to get this merged in?

I think this is great; unless @mattheu has any objections I'll fix the build and get this merged.

Once it's merged, it'd be nice to add some documentation to the [callbacks](https://github.com/wp-shortcake/shortcake/wiki/Event-Attribute-Callbacks) in the wiki, and maybe add your example code for enabling a rich text field as well?

goldenapples:
... merging!

mehigh:
Thank you. I'll be on holidays, and after I get back I'll make the necessary additions to the wiki and point to the rich text field implementation.

mattheu:
Thanks for this. Looks great. 👍

mehigh:
As I returned back to computers-land, I'll pick this one up and add a comment to #236 with the update to the wiki to explain how to achieve a rich text editor using the new hooks... even if it's not a new field, it's a matter of adding a class, which is just as easy.

mehigh:
I've added an update to the wiki here regarding these 3 new hooks, with a small code sample:
https://github.com/wp-shortcake/shortcake/wiki/Event-Attribute-Callbacks
I've also added a link to my implementation of the rich text editing, hope that's all right.

Issue #635, opened 2016-08-19
listItemImage doesn't work for full image path


ajambrovic:
The render function in insertShortcodeListItem creates an HTML element if a dashicon is used, but doesn't do that for full image path - the path gets outputted. 

Processing:

```
var data = this.model.toJSON();
this.$el.attr('data-shortcode', data.shortcode_tag);

if (('listItemImage' in data) && 0 === data.listItemImage.indexOf('dashicons-')) {
    var fakeEl = $('<div />').addClass( 'dashicons' ).addClass( data.listItemImage );
    data.listItemImage = $('<div />').append( fakeEl ).html();
}
this.$el.html(this.template(data));
return this;
```

Template:

```
<div class="add-shortcode-list-item-icon">{{{ data.listItemImage }}}</div>
```

Proposed solution:

```
if ('listItemImage' in data) {
    if (0 === data.listItemImage.indexOf('dashicons-')) {
        var fakeEl = $('<div />').addClass('dashicons').addClass(data.listItemImage);
        data.listItemImage = $('<div />').append(fakeEl).html();
    } else {
        var fakeEl = $('<img src="' + data.listItemImage + '" />').addClass(data.listItemImage);
        data.listItemImage = $('<div />').append(fakeEl).html();
    }
}
```

goldenapples:
Apologies for missing this issue. What I've always done is to pass a full image tag for `listItemImage` - see how it's done in Shortcake Bakery, for example: https://github.com/wp-shortcake/shortcake-bakery/blob/d25d6d3636f7fe5ce9127aa7fd35be85626ebf7a/inc/shortcodes/class-googledocs.php#L12

If you pass anything other than a string which begins with `dashicons-`, it'll just be used as-is for the list item image. I think this approach is much more flexible, as it allows you to use things like svgs or divs. The solution you're proposing would break that existing usage, so we'd need to think it through a bit more.

ajambrovic:
Hi - after reading your and my comment, I realise I could have came to the same conclusion alone - I agree, the way you do it is much more flexible, and the change by Perturbatio reflects what I wanted to suggest, so this is resolved - thanks!

PR #640, opened 2016-09-28
Fix for #635

Perturbatio:
added image tag if the listItemImage param exists but does not start with 'dashicons-'

### DIFF
--- a/js/src/views/insert-shortcode-list-item.js
+++ b/js/src/views/insert-shortcode-list-item.js
@@ -12,12 +12,17 @@ var insertShortcodeListItem = wp.Backbone.View.extend({
 
 	render : function() {
 
-		var data = this.model.toJSON();
+		var data = this.model.toJSON(),
+			fakeEl;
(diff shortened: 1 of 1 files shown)


goldenapples:
I just left a note on [the issue](https://github.com/wp-shortcake/shortcake/issues/635#issuecomment-250209275) - basically, we can already use images by passing the full image tag, so I don't think this approach is necessary. It might be a nice shortcut to have rather than passing the full image tag, but I think it'd need to be thought through a bit more, so that we don't stuff all the markup that people are passing now here into the src of an image tag.

Perturbatio:
I think the main problem then is that the comment in dev.php says that you should use the dashicon-<class> or Full image URL rather than HTML

goldenapples:
> the comment in dev.php says that you should use the dashicon- or Full URL rather than HTML

Oh, that is a bug then. Care to close this and open a PR fixing the erroneous comment in dev.php?

PR #642, opened 2016-09-29
Fix listItemImage documenation in dev plugin

goldenapples:
Fixed error in comment for listItemImage which has led to confusion exhibited in #635 

Rebased on master to avoid merging work from another pull request. Closes #641, props @Perturbatio.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require-dev": {
-        "squizlabs/php_codesniffer": "2.6.2",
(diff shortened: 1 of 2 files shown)


Perturbatio:
sorry I meant to go back and fix a typo in the comment I added "or full HTML **use**" remove use, that's what happens with lack of sleep and a rushed pull request :)

goldenapples:
> remove use, that's what happens with lack of sleep and a rushed pull request :)

Oh thanks, I missed that. Will fix.

PR #644, opened 2016-10-03
Allow percents when using decodeURIComponent for encoded HTML support

### LINKED ISSUE #634: textarea with encode => true and percent sign
So I've added a few textareas through the plugin and all is fine and working great. Until someone uses the percent sign (%) in the textarea.

We're using then encode parameter for the textarea since that's was the only option to get the output we wanted, but it wreaks havoc if someone uses the % sign. Or rather it becomes unsaveable.

I tried pulling the latest files from GitHub as of 2016-08-18 11:37 CET but the error persists.

**In short:**
- Registered a shortcode
- Added Shortcake for that shortcode (see attached file)
- Shortcake is textarea with encode => true
- If a user uses the Shortcake feature and writes text without the % sign (avoiding it, writing percentage instead etc.) all is fine. I have yet to run into a sign/char that triggers the same behaviour.
- If a user inputs % into the textarea the Update button will do nothing.
- Chrome will log the following error in the console: Uncaught URIError: URI malformed ( shortcode-ui.js?ver=0.7.0-alpha:214)

So if I change line 214 in shortcode-ui.js from
`attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ) ) ), { silent: true } );`
to
`attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace("%", "&#37;") ) ), { silent: true } );`

It will update and save. But I am aware it's nothing more than a ugly hack for now and not a viable solution, and the next time you edit the shortcode field the % sign will be replaced by the HTML entity (but it will of course output correctly). I tested it more as a way to verify the culrprit.

A test case for registering the fields but with other settings removed (they do not do any difference):
[testcase.php.zip](https://github.com/wp-shortcake/shortcake/files/424560/testcase.php.zip)

So, how do I/we fix this in a better way?

Edit:
I have only tested this on a textarea input type. But I'm sure it will happen in other input fields as well.

sc0ttkclark:
Fixes #634

Support percent characters in values that get encoded, otherwise JS errors like `Uncaught URIError: URI malformed(…)` happen and stop the shortcode embedding process.

### DIFF
--- a/js/src/models/shortcode.js
+++ b/js/src/models/shortcode.js
@@ -75,7 +75,7 @@ Shortcode = Backbone.Model.extend({
 
 			// Encode textareas incase HTML
 			if ( attr.get( 'encode' ) ) {
-				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( "%", "&#37;" ) ), { silent: true } );
+				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( "%", "&#37;" ) ) ), { silent: true } );
 			}
 
 			attrs.push( attr.get( 'attr' ) + '="' + attr.get( 'value' ) + '"' );
--- a/js/src/models/shortcode.js
+++ b/js/src/models/shortcode.js
@@ -75,7 +75,7 @@ Shortcode = Backbone.Model.extend({
 
 			// Encode textareas incase HTML
 			if ( attr.get( 'encode' ) ) {
-				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ) ) ), { silent: true } );
+				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( "%", "&#37;" ) ), { silent: true } );
 			}
 
 			attrs.push( attr.get( 'attr' ) + '="' + attr.get( 'value' ) + '"' );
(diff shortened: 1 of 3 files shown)


js/src/models/shortcode.js
```diff
@@ -75,7 +75,7 @@ Shortcode = Backbone.Model.extend({
 
 			// Encode textareas incase HTML
 			if ( attr.get( 'encode' ) ) {
-				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ) ) ), { silent: true } );
+				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( "%", "&#37;" ) ), { silent: true } );
```
goldenapples: This line is missing a closing parenthesis.

goldenapples (changes requested):
I don't know why I've never come across this bug - must have been sheer luck, I guess.

Thanks for troubleshooting this and finding a fix. The only problem here is the missing closing paren in the src file - can you fix that, and I'll merge this?

js/src/models/shortcode.js
sc0ttkclark: Fixed :)

goldenapples (approved):

goldenapples:
Thanks @sc0ttkclark and @kosmiq, merging this and will include the fix for version 0.7.0.

cooms13:
Thanks everybody!

Issue #648, opened 2016-10-26
term_select field is not compatible with WP < 4.5


petitphp:
The term_select field use `get_terms( $args )` to retrieve the results for the AJAX request. This break WordPress install before 4.5 where the first argument was a `string|array` of taxonomies's name.

PR #649, opened 2016-10-26
Use pre 4.5 arguments notation for get_terms to maintain compatibility

petitphp:
This should fix issue #648


mattheu:
We discussed this on slack #feature-shortcode. We're unsure about this one. We would rather avoid changing the code to support older versions of WordPress, especially with 4.7 around the corner.

petitphp:
Thanks for the response. I just read the discussion on slack and I understand your concerns. Closing the pull-request and issue.

PR #653, opened 2016-11-08
Use post params to build ajax endpoint rather than string concatenation

goldenapples:
Rather than building the admin-ajax url to post a bulk preview request to, use a post parameter to specify the action.

Adds compatability with WPML and other plugins which filter the admin-ajax url to add additional query string parameters to it. Also, it seems like better practice to use request parameters rather than concatenating strings to build urls.

An alternate approach to #568

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -838,7 +838,8 @@ var Fetcher = (function() {
 			return;
 		}
 
-		var request = $.post( ajaxurl + '?action=bulk_do_shortcode', {
+		var request = $.post( ajaxurl, {
+				action: 'bulk_do_shortcode',
 				queries: _.pluck( fetcher.queries, 'query' )
 			}
 		);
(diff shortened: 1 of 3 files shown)


js-tests/build/specs.js
```diff
@@ -838,7 +838,8 @@ var Fetcher = (function() {
 			return;
 		}
 
-		var request = $.post( ajaxurl + '?action=bulk_do_shortcode', {
+		var request = $.post( ajaxurl, {
+				action: 'bulk_do_shortcode',
```
westonruter: Why not use `wp.ajax.post()` instead?

goldenapples:
ready for review/merge here.

js-tests/build/specs.js
goldenapples: Oh! yeah, that would make sense here...

goldenapples: Updated in 98e0393

mattheu:
👍

PR #605, opened 2016-05-19
SVG support

pravdomil:
Patch to support SVG using https://wordpress.org/plugins/svg-support/

PR #654, opened 2016-11-08
Support svg images for attachment field thumbnail preview

goldenapples:
This is #605, rebased on top of current master and with the build files checked in.


goldenapples:
Merging this, as the code looks good and we decided that no additional explanation was necessary.

PR #655, opened 2016-11-15
646 update to select2 4 bugs

mattheu:
Ran into a couple of bugs in the new Select2 4 implementation

- If you don't select a value, an error is triggered trying to call `join` on `null` when updating. This is because JavaScript being JavaScript, `'object' === typeof null` is true!
- Error when term select field doesn't return any results. I think the processResults callback should always items, even if its empty.


goldenapples:
Ah, javascript. I never knew how to test specifically for a array, since it doesn't return a unique value with `typeof`. `Array.isArray()` is nice to know, filing that away for the future.

Also returning a structured empty array when there are no results seems like a good catch. I didn't notice any errors coming from returning null, but it seems better to be conservative with the return values there....

PR #656, opened 2016-11-15
Select2 v4 styles

mattheu:
The Select2 4 styles are a little off, due to conflicts with the core WordPress styles.

Additionally, I added some styles to make the select2 fields appear more like the default input fields. I personally like this, although maybe we need better indication that these fields are searchable? Maybe just a placeholder text?


goldenapples:
> maybe we need better indication that these fields are searchable? Maybe just a placeholder text?

I'm not exactly sure right yet. By default, all of these Select2 fields will get a placeholder of "Search", which is clear enough when a field has no selections yet. Once there's an item selected in the field, the selected label is a pretty clear indication. I think this is good as a start.

The 300px max width is a little tight for some field types. For example, post titles are often going to be longer than can comfortably fit in that space:

screen shot 2016-11-15 at 9 12 04 am
But we already have classes in place to allow customizing the field styles in cases like these.

The next steps I see with these fields will be making it possible to more easily override the `templateResult` and `templateSelection` methods for a field (for example, to show the featured image, the date, or the content type in the selection). Making that possible will require some more attention to design, so shipping the feature with very minimal styling is a good approach.

mattheu:
Ah ok, maybe make it wider. I just picked 300 to match the other inputs.

mattheu:
Also good ideas on improving the UI with thumbnail. One thing I often find on client sites, is they include multiple post types, and would be good to indicate this too.

mattheu:
Placeholder is actually behaving a little odd. I don't see it at all, but if I unselect all options it does show.

I'll dig into this.

PR #664, opened 2016-11-29
compatibility with WooCommerce 2.6.8 select2

mladjom:


goldenapples (commented):
I don't think that the conditional checking the current post type is necessary. I think we should probably register the script so that it's available anywhere that there might be an editor on the page - it'll only be enqueued when enqueuing editor scripts anyways.

With this conditional in place, any WP editors not on post edit screens will have errors setting up the shortcake scripts.

Issue #662, opened 2016-11-28
Issue with select2


yanknudtskov:
The issue from https://github.com/wp-shortcake/shortcake/issues/646 still persists in version 0.7.0

mladjom:
I was able to solve this by adding 
<pre>$min = '.full'; </pre>
in class-shortcode-ui.php on line 200

goldenapples:
I noticed that... it looks like the only difference between the version of select2 in Customize Posts and here is that Customize Posts is loading Select2.full.js, which includes some back compat shims for methods which were removed from the 3.x branch of select2.

If enqueing that same file here is enough to make the two plugins work well together, that seems like a good fix to make.

mladjom:
It seems that plugin is locking WooCoomerce products edit screen.
[https://wordpress.org/support/topic/version-0-7-0-interferes-with-woocommerce-products](https://wordpress.org/support/topic/version-0-7-0-interferes-with-woocommerce-products)
Even in WooCommerce settings wherever select2 is used select fields are hidden.

mladjom:
I created pull request which at least temporary solves this  #664

yanknudtskov:
awesome @mladjom :-)

PR #666, opened 2016-11-29
Use select2.full.js for compatibility with other plugins which enqueue it

goldenapples:
Built on top of #664.

This adds the check for the SCRIPT_DEBUG constant back in, to load a minified version of select2.full.js unless SCRIPT_DEBUG is true.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -197,12 +197,16 @@ class Shortcode_UI {
 	public function action_admin_enqueue_scripts( $editor_supports ) {
 		add_editor_style( trailingslashit( $this->plugin_url ) . 'css/shortcode-ui-editor-styles.css' );
 
-		$min = '';
+		$min = '.full';
 
-		wp_register_script( 'select2',
-			trailingslashit( $this->plugin_url ) . "lib/select2/js/select2{$min}.js",
-			array( 'jquery', 'jquery-ui-sortable' ), '4.0.3'
-		);
+		$current_post_type = get_post_type();
+		if ( $current_post_type ) {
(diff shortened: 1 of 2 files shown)


goldenapples:
@mladjom Can you check if this works for you with WooCommerce?

mladjom:
It's good to have this check for loading a minified versions. However while WooCommerce product screen is not locked anymore search under the tab linked products is not working nor select fields under the WooCommerce settings. Real problem is that WooCommerce uses 2.5.3 version of select2 and I didn't find out that they are planing on any update to 4.0.x

goldenapples:
Hmm. That does seem to be a bit of a problem.

To hack around the issue, you can always unregister the version of the select2 js and css files that Shortcake registers on those product pages and register the versions which WooCommerce uses you need registers - as you were doing. But doing this would break any shortcode forms that use select2, or any other plugins that use the latest version of the library.

This PR seems to fix the compatibility issue with Customize Posts (#662) for me - if someone else can confirm that, I'd appreciate it.

But I'm not sure what to do about compatibility with plugins like Woo that use an older version of the library.

mladjom:
Yes, issue with Customize Posts is fixed. 
I don't see a good solution for WooCommerce with select2. How about alternative to select2 like chosen?

grappler:
I have a similar issue with CMB which is using an older version. https://github.com/humanmade/Custom-Meta-Boxes/issues/363

I am just going to test the effect of using the full version will report back here. It may be good to add a check `if ( wp_script_is( 'select2' ) ) {` if there is another script is enqueued.

mattheu (approved):
👍

PR #667, opened 2016-11-30
Don’t URL encode output from color picker in dev.php example

isGabe:
See https://github.com/wp-shortcake/shortcake/issues/577#issuecomment-264002615

@goldenapples review/merge?


goldenapples (approved):
Octothorpes are ok by me.

PR #673, opened 2016-12-06
For select fields, set model value on initial render

### LINKED ISSUE #668: Initial value not inserted in shortcode for select fields
Summary: when initially saving inserting a shortcake component, if no fields are changed, none of them are actually inserted into the shortcode. This is an issue since it makes it problematic to save default values.

Screencast example:
dykibkq3ar

Doesn't seem like a huge deal, however this cause problems re: changing defaults later on. For example, right now the first value in the template dropdown in the example above is "Simple", which is essentially the default value. If, however, later on behind the scenes we want to switch our default template behind the scenes, this _will_ affect all components that were only saved once (and hence have no `template` attribute, however it will not affect any component that has been updated at least once. So while the UI looks exactly the same for two saved components, one can actually have the `template` att specified, while the other might not, and thus when the behind-the-scenes default is changed, the two components that look exactly the same from the UI perspective will actually render differently, which is not great.

Is this functionality intended, and if so, why? If not, can we get a fix please?

goldenapples:
Select fields are a bit of a special case as far as attribute fields go, in that the select element's value in the UI can appear to be set without the model value itself being defined. This change handles cases where there's no empty value in the select field, so the first option should function as a default value.

When a select field is rendered, if the model's value is empty and the first element isn't an empty option, this will set the model's value to the value of the option which is visibly selected, to avoid confusion.

Fixes #668

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -1145,12 +1145,15 @@ var editAttributeField = Backbone.View.extend( {
 
 		this.$el.html( this.template( data ) );
 
+		// Ensure default value for select field.
 		if ( 'select' === data.type && '' === this.model.get( 'value' ) ) {
 			var firstVisibleOption = _.first( data.options );
 			if ( 'undefined' !== typeof firstVisibleOption.value ) {
 				this.setValue( firstVisibleOption.value );
 			}
 		}
+
+		this.$el.html( this.template( data ) );
 
 		this.triggerCallbacks();
 
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -1144,6 +1144,14 @@ var editAttributeField = Backbone.View.extend( {
 		}
 
 		this.$el.html( this.template( data ) );
+
+		if ( 'select' === data.type && '' === this.model.get( 'value' ) ) {
+			var firstVisibleOption = _.first( data.options );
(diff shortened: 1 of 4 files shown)


goldenapples:
Here's a fix for #668, ready for review/merge from anyone who wants to take it.

js/build/shortcode-ui.js
```diff
@@ -1144,6 +1144,17 @@ var editAttributeField = Backbone.View.extend( {
...
+
+		// Ensure default value for select field.
+		if ( 'select' === data.type && '' === this.model.get( 'value' ) ) {
+			var firstVisibleOption = _.first( data.options );
+			if ( 'undefined' !== typeof firstVisibleOption.value ) {
+				this.setValue( firstVisibleOption.value );
+			}
+		}
+
+		this.$el.html( this.template( data ) );
```
goldenapples, line 1155: Where's this line coming from? It doesn't seem to be in the `src` file and it duplicates L1146 above...

mattheu:
One edge case I can think of here is if one of the options was an empty string, but this was not the first in the list of options. Would this cause unintended consequences?

mattheu:
Oh oops didn't mean to close!

js/build/shortcode-ui.js
mattheu: recompiled to remove

mattheu:
I think this looks good now!

PR #675, opened 2016-12-15
Handle options with empty values in select fields

goldenapples:
If an option with a null value is present, then let the field value
default to empty, rather than to the first visible option value. Also
sets that value on the data object so that it shows as visibly selected.

Merges into #673


mattheu:
Looks good

Issue #679, opened 2016-12-26
Duplicate wp.media popup dialog markup when add/edit shortcode.


rahulsprajapati:
When we open wp.media model on `.shortcake-add-post-element` button click event or edit the shortcode event we are creating new media model every time. We should either use the existing model if its exist or we should remove existing model. 
Because currently it is adding duplicate markup, like we have one element with id '**inner_content**' for shortcode content which we can't access by id if we have more than one element with same id. 

Here for Edit Shortcode : https://github.com/wp-shortcake/shortcake/blob/master/js/src/utils/shortcode-view-constructor.js#L169-L175
This should be 
```
if ( wp.media.frames.wp_media_frame ) {
	wp.media.frames.wp_media_frame.remove();
}

if ( wp.media.frames.wp_media_new_frame ) {
	wp.media.frames.wp_media_new_frame.remove();
}

var wp_media_frame = wp.media.frames.wp_media_frame = wp.media({
	frame : "post",
	state : 'shortcode-ui',
	currentShortcode : currentShortcode,
});

wp_media_frame.open();
```
OR we can use existing one 
```
if ( wp.media.frames.wp_media_new_frame ) {
	wp.media.frames.wp_media_new_frame.remove();
	wp.media.frames.wp_media_frame = '';
}

if ( wp.media.frames.wp_media_frame ) {
	wp_media_frame = wp.media.frames.wp_media_frame;
} else {
	wp_media_frame = wp.media.frames.wp_media_frame = wp.media({
		frame : "post",
		state : 'shortcode-ui',
		currentShortcode : currentShortcode,
	});
}
wp_media_frame.open();
```
And for `Add Post Element` https://github.com/wp-shortcake/shortcake/blob/master/js/src/shortcode-ui.js#L42-L43
I think we are trying to remove editor by using `wp.media.editor.remove( editor );` but it will not work as the parameter `editor` is target content editor id.
I think this should be :
```
if ( wp.media.frames.wp_media_frame ) {
	wp.media.frames.wp_media_frame.remove();
}
if ( wp.media.frames.wp_media_new_frame ) {
	wp.media.frames.wp_media_new_frame.remove();
}
var wp_media_new_frame = wp.media.frames.wp_media_new_frame = wp.media(options);
wp_media_new_frame.open();
```
OR we can use existing one
```
if ( wp.media.frames.wp_media_frame ) {
	wp.media.frames.wp_media_frame.remove();
[...]

PR #680, opened 2016-12-27
Fix : Duplicate wp.media popup dialog markup when add/edit shortcode.

rahulsprajapati:
Issue: #679

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -1165,13 +1165,25 @@ var shortcodeViewConstructor = {
 		var currentShortcode = this.parseShortcodeString( shortcodeString );
 
 		if ( currentShortcode ) {
+			var wp_media_frame,
+				options = {
+					frame : "post",
+					state : 'shortcode-ui',
+					currentShortcode : currentShortcode
+				};
 
-			var wp_media_frame = wp.media.frames.wp_media_frame = wp.media({
-				frame : "post",
-				state : 'shortcode-ui',
-				currentShortcode : currentShortcode,
-			});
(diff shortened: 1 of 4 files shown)


goldenapples:
This is a big improvement in terms of performance and in terms of avoiding errors caused by duplicate markup.

I'm still seeing one case (well, two - one in addition to the case you pointed out about the "Add Post Element" button) where we will have two media modal models open at a time:

1. Click on an existing shortcode to edit it. Then close it.
2. Click the "Add Media" button and then click the "Insert Post Element" menu item to bring up the same shortcode UI.

In this case, the "Add media" button will set up a second instance of the modal and cause all the same duplicate markup that we've been working on removing.

I'm not quite able to get to the bottom of this right now, but I'll take a look later and see if I can see how we can actually register these frames in a way that works better with the core models in `wp.media`. It looks to me like the shortcode UI view constructor model should probably be registered as a `workflow` in the `wp.media.editor` object using [`wp.media.editor.add()](https://github.com/WordPress/WordPress/blob/c68f1de5f8c1dbd30ba812c4927367d95c6e707a/wp-includes/js/media-editor.js#L817-L913) so that we can open it directly with wp.media.editor.open instead of having to call the lower-level API. I'm not sure about the history of that part of the codebase, so I'm not sure if that approach was explored earlier in the plugin and scrapped, but it looks to me like that's the approach we should be trying to take...

rahulsprajapati:
Yeah, sure I'll also look in this.

goldenapples:
Closing this in favor of #689.

Issue #615, opened 2016-06-07
Insert Element hook


marcin-lawrowski:
It would be great to have a hook that is triggered after clicking "Insert Element" button. There should be a possibility to discard closing of the layer. This hook would be an ideal place to implement the client-side validation.

pySilver:
@marcin-lawrowski any idea how to implement this until this in core? I'm looking for the way to prevent invalid shortcode insertions.

goldenapples:
We'd probably want to insert a filter [in here](https://github.com/rahulsprajapati/shortcake/blob/master/js/src/controllers/media-controller.js#L35-L42), right before the `send_to_editor()` function is called. I'm not sure what context would be useful to pass in, but getting the entire `shortcode` model might be enough. 

What kind of validation would you want to be doing here that couldn't be done on updating each attribute? Are you just thinking of checking to make sure all required attributes are filled out?

pySilver:
@goldenapples in my particular case I want to prevent invalid values to be provided in params since my shortcode talks to an external webservice. In order to make it user friendly I need to make a prior request before inserting shortcode into post and ofc an ability to report an error (field based or form general error)

goldenapples:
Ah, makes sense.

The system of JS hooks we use might not be great for that purpose, since they'd tie up execution while you made your external API call. What if we implemented a JS filter at that spot that could return a Promise as a return value, and then wait until that promise resolves to close the modal and send the shortcode to the editor?

pySilver:
@goldenapples sounds ok :)

I think it would be interesting to have a general validation filter which one can mention in `shortcode_ui_register_for_shortcode` by simply providing callback name. It would be up to developer whether this JS callable is defined and what it does. It's just require to respond with specific format.

We can even scaffold some solution that would tie JS to call some php callback with ajax. Which might be even easier solution for developers.

PR #681, opened 2016-12-29
Add `send_to_editor` filter

goldenapples:
Adds a filter which can be used for client-side validation of a shortcode before sending it to the editor. Basic concept for addressing issue #615; for discussion only at this point. (I think this needs some
kind of syntactic sugar to help work with it, as the API this exposes doesn't follow any established conventions in the WP universe and isn't self-explanatory at all.)

@pySilver @marcin-lawrowski This seems like the bare minimum to make the use case you described in #615 possible. 

Hooking into this and blocking sending the shortcode to the editor would be a matter of returning a rejected promise from that filter, like this:

```
wp.shortcake.hooks.addFilter( 'shortcode-ui.send_to_editor', 
    function( okToSend$, shortcode ) {
        console.log( shortcode );
        okToSend$.reject();
        return okToSend$;
    }
);
```

Any thoughts on making this easier to work with? I like the idea of registering validation methods in the `shortcode_ui_register_for_shortcode` function, but I'm torn as to whether they belong on the attributes, the shortcode as a whole, or both.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -79,9 +79,7 @@ var MediaController = wp.media.controller.State.extend({
 		 * Filter run before a shortcode is sent to the editor. Can be used for
 		 * client-side validation before closing the media modal.
 		 *
-		 * Called as `shortcode_ui.send_to_editor`.
-		 *
-		 *
+		 * Called as `shortcode-ui.send_to_editor`.
 		 *
 		 * @param $.Deferred A promise which is expected to either resolve if
 		 *                   the shortcode can be sent to the editor, or reject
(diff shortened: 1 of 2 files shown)


js/build/shortcode-ui.js
```diff
@@ -70,12 +71,33 @@ var MediaController = wp.media.controller.State.extend({
...
+		 *
+		 * @param $.Deferred A promise which is expected to either resolve if
+		 *                   the shortcode can be sent to the editor, or reject
+		 *                   if not.
+		 * @param Shortcode  The current shortcode model.
+		 */
+		var sendToEditor$ = wp.shortcake.hooks.applyFilters( 'shortcode-ui.send_to_editor', okToInsert$, shortcode );
+
+		// Unless a filter has interfered, resolve the promise, sending the shortcode to the editor.
+		setTimeout( function() { okToInsert$.resolve(true); } );
```
davisshaver, line 92: I'm not that familiar with `$.Deferred.()` and using it with `setTimeout()`, so let me try parroting this back first. If there's a hook applied, `sendToEditor$` could be a rejected promise. Otherwise it should be the value of `okToInsert$`, which is an unfulfilled promise. 

What does the timeout do? Could we set `okToInsert$` as a resolved promise up top and remove the timeout portion? Seems like it might be the same result.

davisshaver:
Re validation, my inclination is that shortcode-level validation would make the most sense and not limit any field-specific use cases.

Issue #564, opened 2015-12-11
Checkbox attribute fields are not set on the model when opening the shortcode UI 


goldenapples:
From https://github.com/wp-shortcake/shortcake-bakery/issues/165

Currently in our shortcode model, it's expected that a checkbox field should store a boolean value.

However, when we [format a shortcode string](https://github.com/wp-shortcake/shortcake/blob/master/js/src/models/shortcode.js#L65-111), we don't treat boolean fields any differently than other field types, and serialize them as quoted values, ie `attribute="true"`.

When opening the shortcode UI a second time for the shortcode, this field is then read as a string, rather than a boolean, and isn't set properly on the shortcode model (callbacks on the field will report its changed value as `"true"`, rather than `true`, for example).

Related issues: #550, #359

Issue #550, opened 2015-11-16
Allow `'value' => 'true' or true or 1` (string, boolean, integer)


cliffordp:
For a `'type' => 'checkbox'` (and possibly other field types) I tested and `'value' => 'true'` works (as a string), but (bool)`true` and (int/bool)`1` do not <em>check the box</em> by default.

It'd be nice if any of these 3 (and their `false` counterparts) would get the job done too.

If not a good idea, please enlighten me.

tyvm

danielbachhuber:
Related #413

PR #682, opened 2016-12-29
Optionally set checkbox values as 'true' or 'false'

### LINKED ISSUE #359: Pass false shortcode value for unchecked checkbox

goldenapples:
Adds the attribute option 'explicitbool' for checkboxes. If a checkbox is registered with 'explicitbool' => true, then the shortcode will be formatted with that attribute string like `attribute="false"` if unchecked. If the shortcode attributes are retrieved using the shortcode tag as the third parameter to `shortcode_atts()`, then this value will be converted back to a boolean automatically.

This allows better handling of default values for attributes - otherwise, if an attribute was set with a default value of "true", then even if it was unchecked, the next time the shortcode was opened to
edit, that default would be applied and the checkbox would be reset to checked.

Fixes #359, #550, and #564. Replaces #413.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -458,7 +458,7 @@ class Shortcode_UI {
 				$out[ $attr['attr'] ] = rawurldecode( $out[ $attr['attr'] ] );
 			}
 
-			if ( $explicitbool && in_array( $out[ $attr['attr'] ], array( 'true', 'false' ) ) ) {
+			if ( $explicitbool && in_array( $out[ $attr['attr'] ], array( 'true', 'false' ), true ) ) {
 				$out[ $attr['attr'] ] = $this->explicitbool( $out[ $attr['attr'] ] );
 			}
 		}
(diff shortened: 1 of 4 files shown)


goldenapples:
Would love to get a review for this, if anyone has the time.

astratagem:
I'd love to see this merged – #359 is still bugging me.

PR #684, opened 2017-01-03
.shortcake-attachment-preview's alignment fix

danielecolangelo:
Attaching several files it may happen that single previews have different heights because of files titles. If you use "float:left" with different heights divs, you could get a messy layout. Using "display: inline-block" with "vertical-align: top" we are sure that divs are aligned regularly.

*bug*

01_upstream

*display: inline-block*
*vertical-align: top*

02_inline-block

*word-wrap: break-word*

03_break-word


goldenapples:
Good fix :100:. Seems to cover all possibilities for this attachment preview.

goldenapples (approved):

PR #685, opened 2017-01-07
Add node-sass as devDependency to fix grunt-sass

westonruter:
Fixes `grunt` builds. Without this I would get:

> Warning: Task "sass" not found. Use --force to continue.
> Error: Task "sass" not found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "grunt-sass": "^1.1.0",
     "grunt-wp-i18n": "^0.5.0",
     "grunt-wp-readme-to-markdown": "~1.0.0",
+    "node-sass": "^4.1.1",
     "remapify": "1.4.3"
   },
   "browserify": {


westonruter:
Build fails due to new i18n sniff doing what it should! https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/742

goldenapples:
This surprises me, because `node-sass` is [in the dependency tree for `grunt-sass`](https://github.com/sindresorhus/grunt-sass/blob/master/package.json#L34) so it should be getting built already.

Also, the `sass` task is provided by `grunt-sass`, so if you're getting the `Task "sass" not found` message, that would seem to indicate that `grunt-sass` hadn't been initialized in the first place.

I'll see if I can reproduce the npm install / grunt problem and merge if so.

goldenapples:
I found out this is required when rebuilding my Linux dev machine this morning (I'm not sure why I never came across it before, other than possibly have node-sass installed globally).

I'll merge this to save other people the trouble of having to discover this.

PR #692, opened 2017-01-26
Re-add deprecated function `editAttributeFieldAttachment.getFromCache`

goldenapples:
Up until #508, we had a public class method exposed on `editAttributeFieldAttachment` called `getFromCache()`, which was being used elsewhere to hook into attachment preview rendering and other useful attachment needs. (For example, the Image Shortcake plugin [uses it to pull alt, caption, and other values from the attachment](https://github.com/wp-shortcake/image-shortcake/blob/ae5170965db0dacd5ce89d6558f26045e1cc3834/assets/js/image-shortcake-admin.js#L57) and prepopulate those fields on the image shortcode.)

I think it would still be helpful to expose this cache, even though it isn't needed anymore internally, just because it costs next to nothing to have in place, and is useful for integration with other fields.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -724,6 +724,8 @@ var editAttributeFieldAttachment = sui.views.editAttributeField.extend( {
 
 		self.initSelection();
 
+		self.currentSelection.on( 'all', this.updateCache );
+
 		self.currentSelection.on( 'all', this.updateValue );
 		self.currentSelection.on( 'add', this._renderPreview );
 		self.currentSelection.on( 'reset', this._renderAll );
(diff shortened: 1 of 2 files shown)


goldenapples:
I'd love to get a review of this, and get this merged back into the plugin.

This seems like the simplest thing I can think to do, to restore the `getFromCache()` method we deleted, but just have it look up the attributes using `wp.media.attachment()` instead of creating its own "cache" storage. It will fix issues like https://github.com/wp-shortcake/image-shortcake/issues/75, which I noticed after updating Shortcake to the latest master.

goldenapples:
Punting on this for now, since I don't see this method being used anywhere other than in Image Shortcake, and it's easier to update that plugin than to maintain a deprecated method here indefinitely.

Issue #683, opened 2016-12-31
Uncaught Error: No select2/compat/initSelection


hirejordansmith:
Hi,

I'm have an issue I can't seem to resolve unless I deactivate the Shortcake (Shortcode UI) plugin and trying to get to the bottom of it.

I am running WP 4.7 with 2016 theme for theme / plugin conflict testing

The only plugins I have installed at the moment are WooCommerce (v2.6.11) and Shortcode UI (0.7.0)

I also removed all of the actual Shortcode UI code so it's just the plugin itself as far as related code.

The problem is that all of the jQuery elements on single product pages and admin pages are breaking so no longer functional.

Any idea what this is related to? I don't see anyone else having the same issue when I review issues.

This is the full error:

Uncaught Error: No select2/compat/initSelection
    at callDep (select2.js?ver=4.0.3:215)
    at req (select2.js?ver=4.0.3:376)
    at select2.js?ver=4.0.3:190
    at Defaults.apply (select2.js?ver=4.0.3:4619)
    at new Options (select2.js?ver=4.0.3:4919)
    at new Select2 (select2.js?ver=4.0.3:5046)
    at HTMLInputElement.<anonymous> (select2.js?ver=4.0.3:5667)
    at Function.each (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,jquery-ui-core,utils,jquery-ui-widget,jquery…:2)
    at a.fn.init.each (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,jquery-ui-core,utils,jquery-ui-widget,jquery…:2)
    at a.fn.init.$.fn.select2 (select2.js?ver=4.0.3:5664)

And it says the error is line 215 in this file 

https://darby.staging.wpengine.com/wp-content/plugins/shortcode-ui/lib/select2/js/select2.js?ver=4.0.3

Any help would be appreciated! I seriously love this plugin! :-)

goldenapples:
I think that problem should be resolved on the master branch here. We have a new release that should probably go out soon to the plugin repository. 

Can you try downloading the plugin from here and see if it works?

hirejordansmith:
Hi Nathaniel,

I downloaded stable version from Master branch from github and see the same issue. Looks like the same version that I'm currently seeing the issue with as well - 0.7.0. Is there a newer version than 0.7.0? Thanks man!

goldenapples:
Hmm. Seems I was wrong in thinking that issue was resolved.

[...]

Issue #670, opened 2016-12-03
Problem with ACF Pro on post_select with multiple => true


raphaelsramos:
After update from 0.6.2 to 0.7.0, all my shortcodes with [ 'type' => 'post_select', 'multiple' => true ] stop working!!!

I've tried removed the 'multiple' => true, but it still dont' work.

goldenapples:
We've had a lot of other people note this issue. It's a conflict between the version of Select2 which we're using here (4.0) and the older version which ACF Pro uses by default.

A solution [noted here](https://github.com/wp-shortcake/shortcake/issues/660#issuecomment-263397486) is to use the `acf/settings/select2_version` filter to ensure that ACF loads version 4 of select2.

kipraske:
There is a problem with the `acf/settings/select2_version` that I was wondering if anyone here has come across. It turns out that using this ACF hook, the ACF select2 custom fields no longer allow the user to deselect all of the options.

Okay, sure this is probably an issue with ACF rather than Shortcake, but maybe someone here has a fix for this already. Also, this will let people know that the above workaround will work for most cases but we still have this bug if you are using ACF fields that use Select2 (Such as the "User" custom field)

PR #702, opened 2017-02-28
Optional compatibility with other Select2 v3 plugins

brandomeniconi:
As discussed in #683 and #670, actually many popular plugin, like Advanced Custom Fields and WooCommerce, still use the Select2 v3 library and make this plugin throw JS errors due to v3/v4 incompatibility.

With this PR the user can: 

`define('SELECT2_NOCONFLICT', true);` in `wp-config.php` 

to make this plugin load select2 4.x in `$.fn.select2v4` namespace, leaving the `$.fn.select2` on v3 for all other plugins. The plugin's JS than uses the `select2v4 ` namespace if exists, otherwise it uses the standard `select2`.

An optional improvement to this PR could be to check if known incompatible plugins  are active with `is_plugin_active()` and automatically set  `SELECT2_NOCONFLICT` to true.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -38,7 +38,7 @@ class Shortcode_UI {
 	private static $instance;
 
 	/**
-	 * Select2 handle
+	 * Select2 library handle.
 	 *
 	 * @access public
 	 * @var string
@@ -67,10 +67,6 @@ class Shortcode_UI {
 		$this->plugin_version = SHORTCODE_UI_VERSION;
 		$this->plugin_dir     = plugin_dir_path( dirname( __FILE__ ) );
 		$this->plugin_url     = plugin_dir_url( dirname( __FILE__ ) );
-
-		if(defined( 'SELECT2_NOCONFLICT' ) && SELECT2_NOCONFLICT){
-			self::$select2_handle = 'select2v4';
-		}
 	}
 
 	/**
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -198,14 +198,14 @@ class Shortcode_UI {
 		add_editor_style( trailingslashit( $this->plugin_url ) . 'css/shortcode-ui-editor-styles.css' );
 
 		$min = ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
-		$noconflict = defined('SELECT2_NOCONFLICT') && SELECT2_NOCONFLICT;
+		$noconflict = defined( 'SELECT2_NOCONFLICT' ) && SELECT2_NOCONFLICT;
 
 		wp_register_script( $noconflict ? 'select2v4' : 'select2',
 			trailingslashit( $this->plugin_url ) . "lib/select2/js/select2.full{$min}.js",
 			array( 'jquery', 'jquery-ui-sortable' ), '4.0.3'
 		);
 
-		if($noconflict){
+		if( $noconflict ){
 			 wp_add_inline_script( 'select2v4', 'var existingSelect2 = jQuery.fn.select2 || null; if (existingSelect2) { delete jQuery.fn.select2; }', 'before' );
 			 wp_add_inline_script( 'select2v4', 'jQuery.fn.select2v4 = jQuery.fn.select2; if (existingSelect2) { delete jQuery.fn.select2; jQuery.fn.select2 = existingSelect2; }', 'after' );
 		}
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -198,13 +198,19 @@ class Shortcode_UI {
 		add_editor_style( trailingslashit( $this->plugin_url ) . 'css/shortcode-ui-editor-styles.css' );
 
 		$min = ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
+		$noconflict = defined('SELECT2_NOCONFLICT') && SELECT2_NOCONFLICT;
 
-		wp_register_script( 'select2',
+		wp_register_script( $noconflict ? 'select2v4' : 'select2',
 			trailingslashit( $this->plugin_url ) . "lib/select2/js/select2.full{$min}.js",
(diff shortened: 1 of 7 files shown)


js/build/shortcode-ui.js
```diff
@@ -1768,7 +1768,8 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 
 		this.preselect( $field );
 
-		var $fieldSelect2 = $field.select2({
+                var select2 = jQuery().select2v4 ? 'select2v4' : 'select2';
```
goldenapples: We could probably pass this handle in along with the localized script variables, so that we don't need to check which name it exists under here.

inc/fields/class-field-post-select.php
```diff
@@ -38,8 +38,10 @@ public function filter_shortcode_ui_fields( $fields ) {
 
 	public function action_enqueue_shortcode_ui() {
 
-		wp_enqueue_script( 'select2' );
-		wp_enqueue_style( 'select2' );
+		$noconflict = defined( 'SELECT2_NOCONFLICT' ) && SELECT2_NOCONFLICT;
+
+		wp_enqueue_script( $noconflict ? 'select2v4' : 'select2' );
+		wp_enqueue_style( $noconflict ? 'select2v4' : 'select2' );
```
goldenapples: One thing that I think might help clean up the code would be to define the Select2 handle in use as a public static variable on the `Shortcode_UI` class. This would get around the need to repeat this pattern of checking the constant in four different places.

goldenapples (changes requested):
Nice, and thanks for the PR! I've tested this out with Woocommerce and can confirm that the select2 functionality from both plugins work as expected. I was concerned that there would also be style conflicts, but it looks as though the classes used have changed enough from select2 v3 to v4 that there aren't any major clashes from loading two separate stylesheets.

> An optional improvement to this PR could be to check if known incompatible plugins are active with `is_plugin_active()` and automatically set `SELECT2_NOCONFLICT` to true.

Yeah, that could work, but I'd prefer not adding that logic right now. This PR should be a good enough functional fix for site owners who come across this conflict, and trying to add magic conflict detection would be adding one more thing to maintain to a plugin which is already undermaintained. We should just add some clearer documentation so that people who encounter this problem don't have to search through issues here to find the solution.

The failing lint has to do with [the relatively new WPCS file name rule](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/806), and doesn't need to be fixed here. I'll open a separate PR to address that.

brandomeniconi:
Thanks for the really nice suggestions, I've updated the code as requested, there is something else that needs to be changed? Should we be more specific for the constants naming something like `SHORTCODE_UI_SELECT2_NOCONFLICT` or `SELECT2_V4_NOCONFLICT` ?

> Yeah, that could work, but I'd prefer not adding that logic right now. 

I agree that magic plugin detection it's unnecessary, I forgot that this plugin is dev-oriented so almost its users are able to define the constant in the config. This can also be done in a separate compatibility plugin if needed.

> We should just add some clearer documentation so that people who encounter this problem don't have to search through issues here to find the solution.

I'm not very good at english (I spend way more time writing PRs than code 😄), can you do that for me? So you can choose the right position in the README. If you want to post it here I'll add it in the PR.

> The failing lint has to do with the relatively new WPCS file name rule, and doesn't need to be fixed here. I'll open a separate PR to address that.

Ok, thanks!

inc/class-shortcode-ui.php
```diff
@@ -199,12 +207,17 @@ public function action_admin_enqueue_scripts( $editor_supports ) {
...
-		wp_register_script( 'select2',
+		wp_register_script( self::$select2_handle,
 			trailingslashit( $this->plugin_url ) . "lib/select2/js/select2.full{$min}.js",
 			array( 'jquery', 'jquery-ui-sortable' ), '4.0.3'
 		);
 
-		wp_register_style( 'select2',
+		if( self::$select2_handle !== 'select2' ){
+			 wp_add_inline_script( self::$select2_handle, 'var existingSelect2 = jQuery.fn.select2 || null; if (existingSelect2) { delete jQuery.fn.select2; }', 'before' );
+			 wp_add_inline_script( self::$select2_handle, 'jQuery.fn.'.self::$select2_handle.' = jQuery.fn.select2; if (existingSelect2) { delete jQuery.fn.select2; jQuery.fn.select2 = existingSelect2; }', 'after' );
```
goldenapples: I suppose the inline variable should be escaped, lest another plugin get us to run some arbitrary javascript here.

goldenapples (commented):
Other than my comment about escaping the variable name we pass into the inline script I like this PR, and it seems very useful for allowing compatability with other plugins. @mattheu @danielbachhuber any objections to merging this so we can cut a version that works with the other plugins that we're conflicting with now? 

I have a separate PR in #703 that fixes our linting issues. I'll have to freshen it up after this is merged, but that should get the build passing again.

inc/class-shortcode-ui.php
goldenapples: ... or, couldn't we just use the handle we've passed in with the localization data? Would `jQuery.fn[ shortcodeUIData.select2_handle ]` work here?

brandomeniconi: Yes of course, great idea and it works! Wordpress loads the localization variables before all scripts, so even if the `shortcodeUIData` is a "translation var" for the `shortcode-ui` script, should be also available to the Select2 script. With that the escaping should be automatically done, right?

goldenapples: > With that the escaping should be automatically done, right?

Yeah, I think the escaping that's being done to prepare the localization strings for json encoding is safe enough. It may not actually guarantee a valid attribute name, but it will prevent passing something like `);alert('xss')` in and running that arbitrary script, which was my concern.

mattheu:
@goldenapples I think this looks good 👍

brandomeniconi:
I've updated the PR with latest requests. Can you write the documentation? 😃

goldenapples (approved):

goldenapples:
This looks great, thanks for coming up with it and working it all through!

I'll merge and write up some documentation on how to enable this.

PR #703, opened 2017-03-01
Fix WPCS class filename and array formatting lints

goldenapples:
Fixes a couple of codesniff issues that are showing up as build errors:

- renames the classes defining the user select, post select, term select, colorpicker and attachment fields to match their filenames
- reformat associative arrays to follow guidelines (one key per line, comma ending each line).

Should get the build passing again (once Travis gets back up).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -73,7 +73,7 @@ module.exports = function( grunt ) {
 				src: './'
 			},
 			options: {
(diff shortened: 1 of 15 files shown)


goldenapples:
Just fixes for a couple new codesniff issues. 

The only thing that _might_ break functionality is the renaming of the `Shortcake_Field_Attachment` and `Shortcake_Field_Color` classes for consistency (to `Shortcode_UI_Field_Attachment` and `Shortcode_UI_Field_Color`). I did a github search to see if any other plugins were accessing these classes by name, and didn't see anything, so it should be safe to rename them.

PR #705, opened 2017-03-04
Reset the currentShortcode value and the menu title on close

goldenapples:
Merges into #689 

We're correctly setting the state of the WP media modal back to
'insert', as used by WP core, when a user closes, submits, or dismisses
the shortcode UI modal. However the currentShortcode state wasn't being
reset, and the modal title wasn't being reset, so the next time the user
opened the media modal, they'd see "{Shortcode} Options" as the menu
item title, rather than "Insert Post Element" as it should be.

This change deletes and resets these props on closing the shortcode ui
frame. It could certainly stand to be refactored, but I didn't see the
best way of doing that offhand.

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -1181,6 +1181,10 @@ var shortcodeViewConstructor = {
 
 			// Make sure to reset state when closed.
 			frame.once( 'close submit', function() {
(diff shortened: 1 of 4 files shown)


mattheu (approved):
Looks good 👍

PR #652, opened 2016-11-08
Select support for multi

mattheu:
As https://github.com/wp-shortcake/shortcake/pull/567/files is touching some code that is changing, I've whipped up a new version of this that builds on the code from https://github.com/wp-shortcake/shortcake/pull/591 and https://github.com/wp-shortcake/shortcake/pull/651

PR #567, opened 2015-12-15
Pre select SELECT field with multiple values

pereirinha:
I came across with a situation where I needed to pre select options in a Select field with multiple values. Perhaps it's easier to understand by inspecting this code:

```
array(
    'label'    => esc_html__( 'Select field', 'reactor' ),
    'attr'     => 'select_field',
    'type'     => 'select',
    'options'  => array( 'First', 'Second', 'Third', 'Fourth', 'Fifth', 'Sixth', 'Seventh', 'Eighth', 'Ninth', 'Tenth' ),
    'value'    => array( 2, 3, 6 ),
    'meta'     => array(
        'multiple' => true,
    ),
    'description' => esc_html__( 'Hold down the Ctrl (windows) / Command (Mac) button to select multiple options.', 'reactor' ),
) 
```

This push request aims to allow this use case while keeping previous functionality.
I think it might be useful for others.

PR #707, opened 2017-03-09
pre-select multi-select options via value param

galengidman:
This was addressed in #567, and stated to be complete in #652, but the current code still doesn't allow for multiple, pre-selected values for multi-selects. This is basically a rip-off of @pereirinha's original PR that allows for an array or comma separated list of values to passed to `value` that will pre select a default set of options.

### DIFF
--- a/inc/templates/edit-form.tpl.php
+++ b/inc/templates/edit-form.tpl.php
@@ -54,7 +54,7 @@
 				<# if ( 'options' in option && 'label' in option ) { #>
 					<optgroup label="{{ option.label }}">
 						<# _.each( option.options, function( optgroupOption ) { #>
-							<option value="{{ optgroupOption.value }}" <# if ( optgroupOption.value === data.value ){ print('selected'); } #>>{{ optgroupOption.label }}</option>
+							<option value="{{ optgroupOption.value }}" <# if ( ! _.isEmpty( _.filter( data.value, function(val) { return val === optgroupOption.value; } ) ) ) { print('selected'); } #>>{{ optgroupOption.label }}</option>
 						<# }); #>
 					</optgroup>
 				<# } else { #>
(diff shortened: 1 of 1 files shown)


goldenapples:
Hmm. You're right that this part of that PR slipped through the cracks when it was rewritten as #652.

I agree with @mattheu that this long ternary is pretty unreadable, but I'd like to get this bug fix in.

goldenapples:
What happens if an option value contains a comma?

I like the part of this which handles an array passed into the attribute `value`, but I think trying to automatically treat a value containing a comma as an array and split it might be making too much of an assumption about the user's intention.

Should we also make sure that `multiple` is true before preselecting multiple options?

galengidman:
Dropping the comma splitting makes a lot of sense... I've pushed that up.

We could check for `multiple` and ensure that `data.value` is a valid array before selecting multiple, but that's a pretty long and messy ternary (at least my attempt at it) for what I wouldn't think would be much of a perf gain.

```js
<# if ( ! _.isEmpty( data.meta.indexOf( 'multiple' ) !== -1 && _.isArray( data.value ) ? _.filter( data.value, function(val) { return val === option.value; } ) : data.value === option.value ) ) { print('selected'); } #>
```

vs

```js
<# if ( ! _.isEmpty( _.filter( data.value, function(val) { return val === option.value; } ) ) ) { print('selected'); } #>
```

goldenapples:
I finally got around to giving this a round of testing.

This works pretty well and the code looks good. Only thing I've found is that the preselected logic doesn't work the same way for options inside an optgroup as it does for options which are outside of an optgroup. Could you apply the same condition to the selected logic [here](https://github.com/galengidman/shortcake/blob/master/inc/templates/edit-form.tpl.php#L57) as you have in the line you changed?

Otherwise, this looks good and will be great to get merged.

galengidman:
That should do ya.

goldenapples (approved):
:100: Looks good to me. Thanks for finishing this one off...

Issue #711, opened 2017-03-27
Explain better in readme, fallback more gracefully


joyously:
I just installed version 0.7.1.
The readme.txt file sort of tells me what to do, but not really. In order to "make using WordPress shortcodes a piece of cake", I would expect that I could see at least a list of registered shortcodes while in the editor. Even if they didn't integrate with this plugin, they have to register, so this plugin should be able to list them.
All I see is a button called "Add Post Element", which to me says nothing about shortcodes. I had to deactivate the plugin to see if the button went away to verify that this plugin put it there. (you can't even mention that button in the readme?)  What I saw in the readme was the description of a screenshot saying to use the "Add Media" button.

So I clicked the Add Post Element button. Nothing happens. That's not a piece of cake. It should say something if no shortcodes are registered (or not even be a button), or give me a list of registered shortcodes. I tried clicking Add Media. I got the standard Media modal with nothing about shortcodes or "Post Elements".

So, no interface at all for the hundreds of existing shortcodes? What good is that?

goldenapples:
I agree that the documentation could be much clearer about what this plugin does. In its current form, this is really just a developer utility plugin that doesn't do much on its own.

If you'd like to be able to use shortcodes off the bat to see how the UI for this plugin functions, I'd try installing either [Shortcake Bakery](https://github.com/wp-shortcake/shortcake-bakery), which enables a number of useful embed scripts as shortcodes, or [JS Widgets](https://github.com/xwp/wp-js-widgets), which provides a shortcode UI and customizer controls for all of the WordPress core widgets.

The problem of providing an editor UI for shortcodes isn't quite as simple as just listing all the shortcodes available - while we could, in theory, show a list of the shortcodes available, there's no way of hinting what options are used, their types or allowed values, etc. That was precisely the problem this plugin was designed to solve by providing an interface to register the allowed attributes for shortcodes and standardize a method of setting them.
   
>  It should say something if no shortcodes are registered (or not even be a button)

Those are both good suggestions. It would be good to

1. show an admin message when the plugin is activated if there are no shortcodes with shortcode ui registered, saying something along the lines of "This plugin does nothing on its own until you have shortcodes with a user interface defined. In order to get the full benefit of shortcode previews, you should ..."
2. Only output the "Add post element" button if there are shortcodes/post elements defined to use.

Would this address your comments?

Also, noted that the readme is way out of date and needs updating with current screenshots and information about the current version.

joyously:
[...]

PR #714, opened 2017-03-29
Add finnish translation for Add Post Element

teropaajoki:


goldenapples:
Thank you! I'll merge this so that it can be in the next release. Hopefully we don't add many new strings from this point on, but as there's discussion of adding new help text and admin notices (as in #711), there may be more changes requiring localization coming.

PR #715, opened 2017-03-29
Communicate warning to users when no shortcodes with UI are registered

goldenapples:
This is a first attempt at addressing some of the issues pointed out in #711.

First off, if this plugin is activated, but no shortcodes are registered with Shortcake UI, an admin notice will be shown:

screenshot from 2017-03-29 13-27-55

> The Shortcode UI plugin will not do anything unless UI is registered for shortcodes through a theme or plugins. For examples, see here.

with a link to [the wiki page here](https://github.com/wp-shortcake/shortcake/wiki/Shortcode-UI-Examples) with plugins which provide supported shortcodes.

Secondly, if there are no supported shortcodes available, the "Add Post Element" button will not be shown, as it doesn't do anything and is a source of potential confusion.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -226,7 +226,7 @@ class Shortcode_UI {
 	 * @return bool
 	 */
 	public function has_shortcodes() {
-		return ( bool ) $this->get_shortcodes();
+		return (bool) $this->get_shortcodes();
 	}
 
 	/**
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -73,11 +73,43 @@ class Shortcode_UI {
 	 * Setup plugin actions.
 	 */
 	private function setup_actions() {
+		add_action( 'admin_notices',             array( $this, 'action_admin_notices' ) );
 		add_action( 'admin_enqueue_scripts',     array( $this, 'action_admin_enqueue_scripts' ) );
 		add_action( 'wp_enqueue_editor',         array( $this, 'action_wp_enqueue_editor' ) );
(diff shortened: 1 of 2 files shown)


goldenapples:
@mattheu or @szepeviktor Care to review this activation notice for readability? It also disables the "Add Post Element" button if there are not post elements with shortcode UI registered, to avoid confusion.

szepeviktor (approved):
I confirm that it is a one-time notice and it is OK.

inc/class-shortcode-ui.php
```diff
@@ -81,6 +82,44 @@ private function setup_actions() {
...
+								'target' => array(),
+							),
+						)
+					),
+					'https://github.com/wp-shortcake/shortcake/wiki/Shortcode-UI-Examples'
+				) .
+				'</p></div>' . "\n";
+		} else {
+			echo '<div class="notice notice-info is-dismissable"><p>' .
+				esc_html__( 'That\'s all! Try out the shortcode UI through the "Add Post element" button in the post edit screen.', 'shortcode-ui' ) .
```
mattheu: I'm not so sure about 'Thats all'. I'm concerned it won't be immediately clear what its referring to. EG if the tab has been left open and the user comes back to it, the context of having just clicked activate is lost. 

How about `Shortcode UI installed. Try it out through the "Add Post element" button in the post edit screen."

'

mattheu (commented):
Looks good. I commented with my thoughts on the notice text - but if you'd like to keep it as you have it - which is maybe a bit more casual, I'm OK with that.

inc/class-shortcode-ui.php
goldenapples: Good point. I'll update.

PR #717, opened 2017-04-04
Add i18n for strings in attachment field template

### LINKED ISSUE #716: Unstralatable string "Thumbnail Details"
In https://github.com/wp-shortcake/shortcake/blob/master/inc/fields/class-shortcode-ui-field-attachment.php#L138 the string is not gettexted, making it untranslatable:

screen shot 2017-04-03 at 14 41 43

Also, and this is a doubt more than an issue, maybe it shouldn't say "*Thumbnail* Details" at all. Since it is consistently referred to as an "attachment" everywhere else, maybe it should read "*Attachment* Details"

### LINKED ISSUE #718: "required" plugin when using Avada
Hey Guys,

When your plugin is installed, it's adding our "required" label to our plugin area like this: http://d.pr/i/7ur3

Why?

goldenapples:
Changes the text of the subhead in this field template from "Thumbnail
Details" to "Attachment Details", which is more consistent with the
surrounding text. Also wraps this field and the "Edit Attachment" link
in gettext calls, so that they can be translated.

Fixes #716.

### DIFF
--- a/inc/fields/class-shortcode-ui-field-attachment.php
+++ b/inc/fields/class-shortcode-ui-field-attachment.php
@@ -135,14 +135,14 @@ class Shortcode_UI_Field_Attachment {
 				</div>
 
 				<div class="thumbnail-details-container has-attachment">
-					<strong>Thumbnail Details</strong>
+					<strong><?php esc_html_e( 'Attachment Details', 'shortcode-ui' ); ?></strong>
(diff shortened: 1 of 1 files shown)


goldenapples:
@zedejose Does this look correct to you?

zedejose:
Ace. Looks perfect.

PR #719, opened 2017-04-17
Percent replacement only does first replacement

cooms13:
The line of code that I changed originally had a replacement for percent signs, but it only replaces the first instance. The change proposed does a regex replacement so that it changes all percents.


goldenapples:
This is a good catch, as the intention back in #644 was definitely to allow any number of percent characters in attributes.

I'll merge this, and we can get it into the next release.

PR #724, opened 2017-04-26
Fix shortcake duplication bug

### LINKED ISSUE #723: Wordpress 4.7.4 edit element issue
Hello

Since updating to wordpress 4.7.4 I am having issues updating post elements. The edit dialog opens and you can make changes however when you click the insert / update button nothing is changed, instead a new element is added to the bottom of the editor and the old one is left untouched.

Steps to reproduce:
1. Add a new element to the page (this seems to be working ok and inserting where it should)
2. Click on the element in the visual editor and click edit
3. Make changes to the element
4. Click the insert / update button
5. The updated element isn't displayed, a new element has been added to the bottom of the page

mattheu:
The bug: since 4.7.4, updating an existing shortcake inserts a duplicate instead of actually updating.

Closes #723 

I figured I'd take a look at the code for core MCE views (galleries etc), to try and work what they do differently and why it is only an issue for shortcake. I noticed that there is an `update` callback passed to the `edit` method of MCE views. But we aren't using this and just use `send_to_editor`. 

To be honest, I'm somewhat surprised this was ever smart enough to know when to update, and when to insert.

This pull request updates the shortcode state controller to allow for a custom insert callback, and defaults to `send_to_editor` if not provided.

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -840,7 +840,7 @@ Shortcode = Backbone.Model.extend({
 
 			// Encode textareas incase HTML
 			if ( attr.get( 'encode' ) ) {
-				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( "%", "&#37;" ) ) ), { silent: true } );
+				attr.set( 'value', encodeURIComponent( decodeURIComponent( attr.get( 'value' ).replace( /%/g, "&#37;" ) ) ), { silent: true } );
 			}
 
 			attrs.push( attr.get( 'attr' ) + '="' + attr.get( 'value' ) + '"' );
@@ -1160,7 +1160,7 @@ var shortcodeViewConstructor = {
 	 *
 	 * @param {string} shortcodeString String representation of the shortcode
 	 */
-	edit: function( shortcodeString ) {
+	edit: function( shortcodeString, update ) {
 
 		var currentShortcode = this.parseShortcodeString( shortcodeString );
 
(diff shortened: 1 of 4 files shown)


goldenapples:
> I'm somewhat surprised this was ever smart enough to know when to update, and when to insert.

I think it was working previously because the shortcode was still considered "selected" in the editor, so send_to_editor just inserted at the cursor point. I still don't fully understand all the changes in TinyMCE, but it looks like the updates from TinyMCE 4.5.0 to use contentEditable=false for blocks in the editor broke the "selection" marker we were depending on to insert over.

This looks like a nice fix.

goldenapples (approved):

rogercoathup:
Guys -- have updated to 0.7.2, but still seeing the duplication issue with WP 4.7.4

I'm seeing a difference in behaviour though between 0.7.1 and 0.7.2:

- With 0.7.1 the duplicate was being inserted immediately after the shortcode being updated.
- With 0.7.2 the duplicate is being inserted at the bottom of the page (below other shortcodes we have on the page)

rogercoathup:
Some additional observations:

The text on the element editor submit button only appears to be correct after the page has been saved:

1. If I add the shortcode, then immediately click to edit it, the button bottom right says "Insert Element", not "Update"
2. If I add the shortcode, save the page, then click to edit the shortcode, the button bottom right says "Update".

On further testing, if I make a change to one of the shortcodes, then hit "update", then edit again, the button resets to "Insert Element", but then goes back to "Update" (I suspect after the auto draft has saved)

rogercoathup:
ignore my comments.

Clearing the browser cache fixed the issues.

PR #728, opened 2017-05-03
Fix state bugs

mattheu:
I ran into a couple of bugs with the modal state.

The first is that the media library / upload tabs show when opening the modal for a second time. This bug has been present since before the last release.

Steps to reproduce:
* Click add post element button
* Close modal with `X` in top right
* Click add post element button again

As can be seen in the screenshot, these tabs should not be visible.

image

The second issue is that after editing an exsisting shortcode, the state is not correctly reset on close.

Steps to reproduce.
* Either insert a shortcode from shortcake or find one that exists.
* Click the edit button
* Click the cross in the top right of the modal
* Click 'add post element' button

You see the following screen. I expect to see the grid of shortcodes I can insert. The media frame state was not correctly reset on close.

image


goldenapples:
This fixes both of the bugs noted. I hadn't even noticed the router showing up on the shortcode frame, but that has been an issue for a long time now... thanks!

The only question I have, in looking over the code, is whether there's any reason that `MediaController.reset()` and `MediaController.resetState()` need to be kept as separate methods. It looks like with one exception, whenever we call one, we're also calling the other one. And [here](https://github.com/wp-shortcake/shortcake/blob/4fdb433ad38a3f867771532222fd09916400f699/js/src/views/media-frame.js#L48), where we're setting the media controller before rendering the insert frame of the window, it would make just as much sense to reset the state as well.

If it makes sense to keep the two actions separated, that's totally fine. I'm just wondering if they can be combined into one method for simplicity though.

mattheu:
> The only question I have, in looking over the code, is whether there's any reason that MediaController.reset() and MediaController.resetState() need to be kept as separate methods.

Agreed. Not sure why I did it like that :)

mattheu:
@goldenapples Updated this to consolidate the 2 methods.

goldenapples (approved):
Looks and works well for me. :+1:

goldenapples:
Wait a sec - I spoke too soon. It looks like the specs.js file isn't updated in the last build commit. I'm not sure why that is. If you can rebuild it here, that'd be great. Otherwise, we can just build it on master...

mattheu:
>  specs.js file isn't updated 

This file is a little out of date and I see a few changes unrelated to this change. I will open a new PR to recompile everything as I"m also seeing CSS changes.

goldenapples:
@mattheu Can you merge in the latest master and finish off this PR, so we can push this fix in a new patch release? If you're busy right now, I can finish it up for you... just let me know.

mattheu:
Will do. I had forgotten this one wasn't merged.

mattheu:
@goldenapples Merged with master - ready to go.

goldenapples:
Nice. I checked in the updated specs file to make sure that tests still pass - they do. I'll merge this now.

PR #729, opened 2017-05-03
Make sure `render_edit` hook fires after media frame is created

mattheu:
Following on from [this ticket](https://github.com/wp-shortcake/shortcake/pull/689#issuecomment-284098837), I have been talking to @rahulsprajapati about how best to achive what he wanted.

Using `setTimeout` as you suggested works fine, however I think I agree with him that there could be a better solution that guarantees that his code is run at the correct time. Having another action `shortcode-ui.after_set_action_update` would be useful.

We currently have an action`shortcode-ui.render_edit`, but shortcake renders the modal content inside a `defer` callback. Because of this I think it makes sense to add an extra action.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -145,8 +145,7 @@ var MediaController = wp.media.controller.State.extend({
 			 *           Reference to the shortcode model used in this overlay.
 			 */
 			var hookName = 'shortcode-ui.render_edit';
-			var shortcodeModel = this.shortcodeModel;
-			wp.shortcake.hooks.doAction( hookName, shortcodeModel );
+			wp.shortcake.hooks.doAction( hookName, currentShortcode );
 
 		}.bind( this ) );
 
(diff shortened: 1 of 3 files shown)


goldenapples:
Oh, this makes a lot of sense.

I'm wondering if it makes more sense just to move the existing `render_edit` hook into the `_.defer` callback, so that functions as expected? Currently that hook fires at an indeterminate time which could be before or after the frame is available. Nailing it down to fire once the frame was rendered seems like it would be the desired behavior in any case I can think of.

mattheu:
Moving the current action rather than a new action also makes a lot of sense. I have updated.

goldenapples (changes requested):
I notice that the build files weren't updated after this change. Can you grunt and check in again?

After running the browserify build, I was able to confirm that after this change, the action got the same context object as before. I think this will work for the rich text field and any other use cases people have for this hook.

```
wp.shortcake.hooks.addAction( 'shortcode-ui.render_edit', console.log );
``` 
 
Looks good to me to merge once the build files are checked in.

PR #730, opened 2017-05-08
updated norwegian translation

mfolkeseth:
- Added missing translations
- Improved old translations
- updated source code references


szepeviktor:
Thank you.
I think no one will check it :)

(There is a [composer problem](https://travis-ci.org/wp-shortcake/shortcake/jobs/228652585) which is irrelevant)

PR #734, opened 2017-05-13
Render thumbnails for non-image attachments

paulsowden:
This is similar to how the wordpress media browser renders attachments; if a non-image attachment has specified a thumbnail meta item use it as the thumbnail.

https://github.com/WordPress/WordPress/blob/31d4d81039226aa8d3f5229af874ee5b9e2a9482/wp-includes/media-template.php#L462


goldenapples (approved):
I like this. It works really well with the auto-generated image previews for PDF uploads added in 4.7 (https://core.trac.wordpress.org/changeset/38949), and should work for other file formats as well.

Unless anyone else sees any issues, I'll merge this over the next few days.

Issue #627, opened 2016-07-20
Image for loaded data for 'post_select' field


mralessio:
Started using Shortcake for my theme, Thank you for your plugin, it levels up a user expirience on the next level.
It'd be really great to get the thumbnail image when loading data for 'post_select' field. That would be very helpful especially in case of lots of posts. Select2 3.5 or 4 has built-in this feature.

Example:
![select2example](https://cloud.githubusercontent.com/assets/16015111/16981395/7ec3ea6c-4e7b-11e6-8668-5be081f973c5.png)

westonruter:
FYI: Inclusion of featured image in a post selector has been implemented in an [Object Selector control for the Customizer](https://github.com/xwp/wp-customize-object-selector).

westonruter:
It's also now in a PR for Customize Posts: https://github.com/xwp/wp-customize-posts/pull/196

<img width="303" alt="jump-to-page-2" src="https://cloud.githubusercontent.com/assets/134745/16999225/c66490ce-4e71-11e6-8a51-d210c68b2484.png">

PR #736, opened 2017-05-15
Show post type in post text

### LINKED ISSUE #735: Show post type in post_select field
Related to #627

I have been using `post_select` for a client and they we were querying posts from two post types. The posts in both post types had the same titles. There was no way to differentiate between the two posts.

image

[Customize Object Selector](https://github.com/xwp/wp-customize-object-selector) has a feature where it shows the post type.
image

Beitrag = Post
Inhaltselement = Content Block (CPT)

grappler:
fixes #735

### DIFF
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -136,8 +136,8 @@ class Shortcode_UI_Field_Post_Select {
 		}
 
 		$query = new WP_Query( $query_args );
-
-		$is_multiple_post_types = count( $query->get( 'post_type' ) ) > 1;
+		$post_types = $query->get( 'post_type' );
+		$is_multiple_post_types = count( $post_types ) > 1 || 'any' === $post_types;
 
 		foreach ( $query->posts as $post_id ) {
 			$post_type     = get_post_type( $post_id );
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -141,7 +141,7 @@ class Shortcode_UI_Field_Post_Select {
 
 		foreach ( $query->posts as $post_id ) {
 			$post_type     = get_post_type( $post_id );
-			$post_type_obj = get_post_type_object( $post_type);
+			$post_type_obj = get_post_type_object( $post_type );
 
 			$text = html_entity_decode( get_the_title( $post_id ) );
 
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -137,11 +137,21 @@ class Shortcode_UI_Field_Post_Select {
 
 		$query = new WP_Query( $query_args );
 
+		$is_multiple_post_types = count( $query->get( 'post_type' ) ) > 1;
+
 		foreach ( $query->posts as $post_id ) {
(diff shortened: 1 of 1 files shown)


inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -137,11 +137,21 @@ public function action_wp_ajax_shortcode_ui_post_field() {
 
 		$query = new WP_Query( $query_args );
 
+		$is_multiple_post_types = count( $query->get( 'post_type' ) ) > 1;
+
 		foreach ( $query->posts as $post_id ) {
+			$post_type     = get_post_type( $post_id );
+			$post_type_obj = get_post_type_object( $post_type);
```
szepeviktor: Please follow WPCS in this line.

inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -137,11 +137,21 @@ public function action_wp_ajax_shortcode_ui_post_field() {
...
+
+			$text = html_entity_decode( get_the_title( $post_id ) );
+
+			if ( $is_multiple_post_types && $post_type_obj ) {
+				$text .= sprintf( ' (%1$s)', $post_type_obj->labels->singular_name );
```
szepeviktor, line 149: Maybe somehow escape `singular_name`?
Although we have json encoding at the end.

inc/fields/class-shortcode-ui-field-post-select.php
goldenapples: This line is triggering a code sniff violation: `Expected 1 spaces before closing bracket`.

goldenapples (commented):
I like this and would support merging once if the one whitespace issue I noted is cleaned up.

It'd be good to tackle providing a better way to template the results for post selection (ie adding the featured image as suggested in #627), but that doesn't need to be done here.

goldenapples (approved):
:+1: Anyone else have suggestions on this? Otherwise, I can merge...

inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -137,11 +137,21 @@ public function action_wp_ajax_shortcode_ui_post_field() {
 
 		$query = new WP_Query( $query_args );
 
+		$is_multiple_post_types = count( $query->get( 'post_type' ) ) > 1;
```
rahulsprajapati: Maybe we should consider 'post_type' => 'any' value.

grappler: I think the `post_type`  should be left to when setting up the setting. For a certain dropdown I may only want posts and in another just pages.

The code is just counting the number of different types of post types in the query results.

rahulsprajapati: yes, but if we use 'post_type' => 'any' that means in this query we will get all post types posts. And still the `count( $query->get( 'post_type' ) )` will be 1. And at that time we won't get post type name in the select dropdown. As we are using this condition.
https://codex.wordpress.org/Class_Reference/WP_Query#Type_Parameters

grappler: Ah, I get you now. The last commit should fix that.

mattheu (approved):

szepeviktor (approved):

grappler:
Can this be merged in? Three people have approved the PR.

mattheu:
👍

PR #737, opened 2017-05-18
Fixed issue with menu click event not being delegated

mviggiano:
The Post Media Frame's click event was not being delegated properly. This was causing the menu to not respond to clicks and therefore not drop down in narrow-viewport scenarios. The issue was being caused by code in media-frame.js returning a function rather than an object for the events property of the mediaFrame object.

Most of the tweaks in this PR aren't mine - they appeared to be preexisting differences between the src and build that were compiled to build by my running of grunt (probably a version difference). The only changes I made were to the events property - the last two diffs.


davisshaver:
Hey @mviggiano! Thanks for the contribution. I was able to recreate the branch with current `npm install` and I tested it locally so I think we are good to merge. Good fix!

szepeviktor:
What is the relation of this repo to alleyinteractive/shortcake?

PR #738, opened 2017-05-22
Ensure select's data value is an array before filtering

jeremyfelt:
During work in #707 to improve support for multi-select fields, handling of comma separated strings was removed as it may have been confusing if an option value had commas in it.

See 9894162ab8bf02941e9611766d5cc3d63d337e96

After this change, string values for single select elements are filtered as if they were arrays. This causes only the first character of a string to be compared to the select value. This in turn results in all select fields set to first option because nothing is seen as selected.

We can accept arrays, but also `split()` with no separator to turn strings into an array so that everything is parsed the same.

### DIFF
--- a/inc/templates/edit-form.tpl.php
+++ b/inc/templates/edit-form.tpl.php
@@ -54,11 +54,11 @@
 				<# if ( 'options' in option && 'label' in option ) { #>
 					<optgroup label="{{ option.label }}">
 						<# _.each( option.options, function( optgroupOption ) { #>
-							<option value="{{ optgroupOption.value }}" <# if ( ! _.isEmpty( _.filter( data.value, function(val) { return val === optgroupOption.value; } ) ) ) { print('selected'); } #>>{{ optgroupOption.label }}</option>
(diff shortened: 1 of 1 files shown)


goldenapples (approved):
Very cool. I never knew that String.split() with no delimiter would behave like that.

There's now even more logic in this template than before, which I was already thinking was getting unreadable. But it works and fixes a bug that has been noted by a few people, so I think we should merge it.

mattheu (approved):

cowanr:
This fix still leaves multi selects not being selected when editing a shortcode.  The short term fix for me was to just ensure that I don't have comma's in my values and split on comma instead of an empty split.
   
`<option value="{{ option.value }}" <# if ( ! _.isEmpty( _.filter( _.isArray( data.value ) ? data.value : data.value.split(","), function(val) { return val === option.value; } ) ) ) { print('selected'); } #>>{{ option.label }}</option>`

kaushik-news-co-uk (approved):

mattheu:
I tried to reproduce the issue @cowanr was having and I think can reproduce. I think they are correct that simply calling `.split(',')` will work instead.

Also I think we can simplify this code by using `_.contains`. I'm going to open a new PR with these additions.

goldenapples:
Closing this PR in favor of #745.

Issue #625, opened 2016-07-05
Exclude self


szepeviktor:
I am developing an ultra-simplified page builder. It works by a CPT called `blokk`.
There is a shortcake which lets you pick 1 blokk by `post_select` on posts and pages **and blokks also**. So a blokk could contain another blokk.

Could we add `$query_args['exclude_self']` to `Shortcode_UI_Field_Post_Select` to prevent selecting the blokk itself?

szepeviktor:
I was trying with `post__not_in` arg but there's no way to tell the post ID at `shortcode_ui_register_for_shortcode()`.

szepeviktor:
Thinking further the current post ID is not in the AJAX request. I think it'll be difficult.
Maybe the solution could be in `edit-attribute-field-post-select.js`?

Issue #661, opened 2016-11-25
render_shortcode_for_preview when post_id is null


phh:
Hi there,

When using Gravityforms and adding a post element to the confirmation editor I'm getting a "Something's rotten in the state of Denmark" error.

My current fix is to filter into map_meta_cap and allow this if $post_id is null.

I guess there should either be a check if it's the WP editor in the edit view or allow the shortcode preview if no post_id is sent.

I've just updated to the latest version of shortcode-ui. The 0.6.2 version had the same issue.

Suggestions?

goldenapples:
Yeah, I've run across that same issue before - see #599.

It's not going to be easy to anticipate all the possible conditions where a user might want to use shortcode previews in editors and define the right capability checks to handle them inside Shortcake. But maybe we could have a filter around the check for whether the user is allowed to preview, and let that be extended elsewhere.

Maybe if we localized the value of `current_screen()` and sent that along with every preview request, that could provide enough context for users to determine whether to allow previews in a given context? I'm thinking changing this simple check [here](https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L363-L365) to something more like

```
$can_preview = current_user_can( 'edit_post', $post_id );

if ( false === apply_filters( 'shortcode_ui_allow_preview_capability', $can_preview, $current_screen_context ) ) {
    return esc_html__( "Something's rotten in the state of Denmark", 'shortcode-ui' );
}
```

would be enough to hook into.

Think that would work here?

phh:
I think a filter would work. And having the current_screen in the filter it would give us alot of flexibility.

But I still think it's sad that the shortcode-ui plugin will not just work "out-of-the-box" with some plugins unless they add the capability check filter.

What is the exact reason for this capability check? It's not like the user is able to insert data from inside this ajax call as it just renders the HTML for the shortcode. Is check capability check really necessary?

goldenapples:
> Is this capability check really necessary?

I think the idea here is just to mirror the capability required to preview a draft post, as a shortcode preview is essentially part of a draft post preview. A user who can't edit a draft post will get a 404 when trying to preview that draft; the capability check just makes that same behavior explicit for the shortcode preview.

[...]

PR #744, opened 2017-06-15
Allow to filter query_args in post_select ajax call

ivankristianto:
I add post id to the post_select ajax call, then a allow to alter the query args for more extendability.

This will solve my issue that return the post list based on which post type it's currently editing.
and it also can fix this issue #625

### DIFF
--- a/js/src/views/select2-field.js
+++ b/js/src/views/select2-field.js
@@ -43,7 +43,8 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 			var request = {
 				include   : _preselected,
 				shortcode : this.shortcode.get( 'shortcode_tag'),
-				attr      : this.model.get( 'attr' )
+				attr      : this.model.get( 'attr' ),
+				postid    : $( '#post_ID' ).val()
 			};
 
 			$.get( ajaxurl, $.extend( request, this.ajaxData ),
--- a/js/src/views/select2-field.js
+++ b/js/src/views/select2-field.js
@@ -43,8 +43,7 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 			var request = {
 				include   : _preselected,
 				shortcode : this.shortcode.get( 'shortcode_tag'),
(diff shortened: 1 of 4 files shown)


goldenapples:
I can see how this would be useful, but - since editors containing shortcodes can appear in more places than just on a post edit screen - I think we might need to figure out a more abstract way of representing the context of the current editor, which then can be passed into queries like ajax call for 'post_select' fields.
 
I have a branch started locally trying to address this for #661 which I need to polish up. I'll see if it can work well with what you have here.

Finally, it might be good to see the javascript logic live inside the parent class [`editAttributeSelect2Field`](https://github.com/wp-shortcake/shortcake/blob/b210b40af132e18269f91f7be7cb3d242901cb94/js/src/views/select2-field.js#L115-L119), rather than in `editAttributeFieldPostSelect`, so that we can make use of the same filters for user select and term select fields?

ivankristianto:
Thanks @goldenapples interested to see how you did the abstraction.
And yes, I agree with you to move the postId arg to parent class. PR updated.

js/src/views/select2-field.js
```diff
@@ -43,7 +43,8 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 			var request = {
 				include   : _preselected,
 				shortcode : this.shortcode.get( 'shortcode_tag'),
-				attr      : this.model.get( 'attr' )
+				attr      : this.model.get( 'attr' ),
+				postid    : $( '#post_ID' ).val()
```
goldenapples: There's actually one other place in select2-fields.js where we should be passing the post ID. Can we also add the `postid` argument to the search ajax request [here](   https://github.com/ivankristianto/shortcake/blob/5632cb3a1159702d7b224990a3e69f0ca97242b6/js/src/views/select2-field.js#L116-L121)?

(Sorry, I think I highlighted the wrong array in my earlier comment - this is the request for initially loading the preselected posts when opening a shortcode ui.)

ivankristianto:
@goldenapples lol i didn't realize that either. I updated the PR

PR #745, opened 2017-06-19
Fix select issue

mattheu:
This builds on, and replaces the work done by @jeremyfelt in https://github.com/wp-shortcake/shortcake/pull/738. Also, thanks to @cowanr for the feedback.

This should handle both standard and multi select fields, and the code is a bit simpler too.


goldenapples (approved):
I've tested this with both regular and multiple select fields, and it works with both (assuming, of course, that the option values don't include commas).

Also `_.contains()` is much easier to read than the way we were filtering this previously, so I like it.

PR #746, opened 2017-06-20
Set active editor

mattheu:
I have run into some issues when working with multiple instances of the wp editor on a page. This PR aims to track the active editor so that when you insert the shortcode it is inserted into the correct editor.

Hold off reviewing/merging this one for now - it was a quick fix for a bug we had and I wanted to track it somewhere. I'll try and put together a some better instructions to reproduce. We ran into this because we have tinyMCE field inside the shortcake edit screen.


rahulsprajapati:
@mattheu  is there something we need to work on this PR, can we merge this to master or add it to next release? 
cc: @goldenapples

mattheu:
Ah yeah - i'll dust this off. We're running this branch so I forgot it wasn't wrapped up :)

js/src/shortcode-ui.js
```diff
@@ -30,7 +30,8 @@ $(document).ready(function(){
 			options = {
 				frame: 'post',
 				state: 'shortcode-ui',
-				title: shortcodeUIData.strings.media_frame_title
+				title: shortcodeUIData.strings.media_frame_title,
+				editor: this.dataset.editor
```
goldenapples, line 34: Just out of curiosity, why do you use `this.dataset.editor` down here, and `$(event.currentTarget).data('editor')` above? Is there any case where the two would be different?

goldenapples (approved):
Looks good to me and works in every case I've tried with multiple WP editors on a page.

goldenapples:
@mattheu Hey, is this ready to merge for 0.7.3? It works for me, do you feel good about it?

goldenapples:
:ship:

PR #749, opened 2017-06-28
Don't override global select2 z-indexing

### LINKED ISSUE #748: Conflict with Kirki plugin
When Shortcake is active, all the select fields in [Kirki](https://github.com/aristath/kirki) stop working(in the customizer). The `SELECT2_NOCONFLICT` constant doesn't help.

#### Versions used:
_Shortcake 0.7.2, latest master branch
Kirki 3.0.3_

goldenapples:
Because the select2 fields we're using here need to show up just over the absolutely positioned media modal, we're setting a z-index value of 16001, just above that modal.

However, there are other places in the WP admin (like the customizer) which require a much higher z-index value, so it would be better to target only the select2 instance which we're using.

This commit uses the "theme" argument to assign an additional classname to the select2 container, and then targets the container in CSS using the new classname.

Fixes #748

### DIFF
--- a/js/src/views/select2-field.js
+++ b/js/src/views/select2-field.js
@@ -106,6 +106,7 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 		var $fieldSelect2 = $field[ shortcodeUIData.select2_handle ]({
 			placeholder: "Search",
 			multiple: this.model.get( 'multiple' ),
+			theme: $.fn.select2.defaults.defaults.theme + ' select2-container--shortcake',
 
 			ajax: {
 				url: ajaxurl,
--- a/css/sass/_select2-fields.scss
+++ b/css/sass/_select2-fields.scss
@@ -40,7 +40,7 @@
 
 }
 
-.select2-container {
+.select2-container--shortcake {
 	z-index: 160000; // Make sure Select2 UI is visible above media modal
 	max-width: 300px;
 }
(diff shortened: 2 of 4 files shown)


js/src/views/select2-field.js
```diff
@@ -106,6 +106,7 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 		var $fieldSelect2 = $field[ shortcodeUIData.select2_handle ]({
 			placeholder: "Search",
 			multiple: this.model.get( 'multiple' ),
+			theme: $.fn.select2.defaults.defaults.theme + ' select2-container--shortcake',
```
goldenapples, line 109: The "theme" parameter gets appended to the string `'select2-container--'` and used as a classname. The default "theme" is "default", so this results in the select2 container getting the classes `'select2-container--default select2-container--shortcake'`. It feels a bit hacky, but it works, and there doesn't seem to be a more straightforward way of targetting only the instance we want, while still allowing it to be styled by the global select2 default css...

goldenapples:
Oh, wait. I think there's a much less hacky way of doing this. 

While investigating the select2 options to fix an unrelated bug, I came across the 'dropdownParent' parameter, which should make it possible for the dropdown to be attached to the media modal, in which case it can just be targetted with a child selector, like `.edit-shortcode-form .select2-container`.

I'll test and see if that works. If so, it seems better than this approach of passing multiple space-separated classes through a parameter that only expects a single slug and exploiting the fact that string concatenation is being used under the hood without any data validation.

PR #750, opened 2017-06-29
Use select2 dropdownParent attr for css targetting

goldenapples:
When interacting with a select2 field, this attaches the dropdown container as a child of the attribute field view, using the `dropdownParent` option in select2.

This allows us to target the specific fields that we need to style with css using parent selectors, without resorting to hacks like trying to set a classname on the container element when it's created.

Replaces #749, as a fix for #748.


szepeviktor (approved):
Well executed.

PR #762, opened 2017-08-11
Fix field type for term_select field in example shortcode

goldenapples:
We added a term select field to the example plugin, but because the type
was set wrong, it isn't actually rendering. This just fixes the field
type, so that it's correct in our documentation.

See #759


dev.php
```diff
@@ -199,7 +199,7 @@ function shortcode_ui_dev_advanced_example() {
 		array(
 			'label'    => __( 'Select Tag', 'shortcode-ui-example', 'shortcode-ui' ),
 			'attr'     => 'tag',
-			'type'     => 'tag_select',
+			'type'     => 'term_select',
```
davisshaver, line 202: 🏷  👍

PR #763, opened 2017-08-11
Avoid saving "null" values on empty select2 fields

### LINKED ISSUE #758: Term Select: Unable to remove term only replace
When using `term_select` with `'multiple' => false,` there is no option to remove the selected term just to override it.

When using `'multiple' => true,` it is possible to remove the term but the value is then defined as null.

It would be good to be able to select "All terms".

This may apply to `post_select` too.

goldenapples:
We're coercing the value of select2 fields to a string, in order to
handle some edge cases in post and term select fields with multiple=true.

However, after deleting all values from a multiple select field which
had been previously updated with one or more values, that attribute's
value is `null`.

We don't want to coerce this empty value to the string "null".
This PR fixes the issue by replacing null attribute values with the
empty string, which doesn't get saved when the model is run through
`Shortcode.formatShortcode()`.

Fixes #758.

### DIFF
--- a/js-tests/build/specs.js
+++ b/js-tests/build/specs.js
@@ -833,8 +833,9 @@ Shortcode = Backbone.Model.extend({
 
 		this.get( 'attrs' ).each( function( attr ) {
 
+			console.log( attr.get('attr'), attr.get('value') );
 			// Skip empty attributes.
-			if ( ! attr.get( 'value' ) ||  attr.get( 'value' ).length < 1 ) {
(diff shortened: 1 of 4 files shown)


goldenapples:
Can I get a review of this bugfix for the next release?

davisshaver (approved):
Looks good! 🔨

goldenapples:
Thanks! Merging this now, will get ready to cut a release once #746 is in.

PR #766, opened 2017-08-24
Add jasmine test for decoding multiple % characters

goldenapples:
Tests that an encoded attribute value with more than one percent
character in it will get decoded properly.

See #756 - this test should pass after that fix is merged.


goldenapples:
Since this PR is only tests and the tests pass, I'll merge as is.

PR #769, opened 2017-08-26
Adds `shortcode_ui_preview` filter

joseadrian:
Added `shortcode_ui_preview` filter to give the option to modify the preview in the WYSIWYG.

Some people want to have the option not to preview the content parsed in the editor because they might break the content, or use the dashicons instead. 

The filter added will allow them to do that: 

```php 
// $options = Shortcode UI Options
add_filter('shortcode_ui_preview', function($return, $shortcode_tag, $shortcode, $options) { 
    if($shortcode_tag == 'special_shortcode_tag') {
         return '<img src="/path/to/image">';
    }

    return sprintf( '<span class="dashicons %s"></span> %s', $options['listItemImage'], $options['label']);
}, 10, 4);
```

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -440,6 +440,25 @@ class Shortcode_UI {
 			// @codingStandardsIgnoreEnd
 		}
 
+		// Get the Shortcode tag and set the response if any
+		preg_match_all( '@\[([^<>&/\[\]\x00-\x20=]++)@', $shortcode, $matches );
+		$shortcode_tag = isset( $matches[1][0] ) ? $matches[1][0] : '';
+
+		/**
+		 * Filter the shortcode preview
+		 *
+		 * Used to render a custom preview for the shortcodes on the editor
+		 *
+		 * @param string $pre_option    The default value to return if the custom preview hasn't been set.
+		 * @param string $shortcode_tag Shortcode tag name.
+		 * @param string $shortcode     Shortcode used.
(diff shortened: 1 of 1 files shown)


davisshaver:
Hi @joseadrian, thanks for contributing. The filter seems reasonable to me. I'm a little suspicious of the new regex in there though. Would you be willing to move that into a utility function and add some test coverage to make sure it handles different shortcode configurations as you expect? Here's the chunk of code I'm looking at.

```
+		preg_match_all( '@\[([^<>&/\[\]\x00-\x20=]++)@', $shortcode, $matches );
+		$shortcode_tag = isset( $matches[1][0] ) ? $matches[1][0] : '';
```

goldenapples:
I was actually wondering if there wasn't a simpler way of doing this. We've talked about a few options that might work:

1. Adding an argument to the `shortcode_ui_register_for_shortcode` args, like `'preview' => false`, and, if that option is set, just showing the text of the shortcode in an mce view, rather than rendering the preview.
2. Using the `SHORTCODE_UI_DOING_PREVIEW` constant to return an alternate rendition of a shortcode when rendering a shortcode preview for the editor. (This is already possible, by filtering `pre_do_shortcode_tag` or `do_shortcode_tag` and checking for the value of the `SHORTCODE_UI_DOING_PREVIEW` constant. But we could probably expose a utility method to make it easier.)

Would either, or a combination of these approaches fit your use case as well?

PR #772, opened 2017-09-05
Remove PHP 5.2 from Travis matrix

goldenapples:
There's no reason we need to be testing against PHP 5.2, and since it's
not pre-installed on Travis any more, it's just causing our tests to
error out. Removing that from the matrix and adding PHP 7.1, which is
actually relevant.


davisshaver (approved):
♻️

PR #774, opened 2017-09-25
Differentiate selected terms by slug and taxonomy

lvl99:
I added in a way to differentiate terms by slug and taxonomy, much like the post_select has a way to differentiate by post type. I often have a number of categories with the same name (e.g. same named categories for men's and women's products on sale) so required the further differentiation.

Additionally, `wp_count_terms` uses `get_terms` anyway, so I edited it to work better with IDEs like PHPStorm which complain about `wp_count_terms` requiring `$taxonomy_type` to be a string.

### DIFF
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -143,11 +143,12 @@ class Shortcode_UI_Field_Post_Select {
 			$post_type     = get_post_type( $post_id );
 			$post_type_obj = get_post_type_object( $post_type );
 
-			$text = html_entity_decode( get_the_title( $post_id ) . ' - ' . get_post_field( 'post_name', $post_id ) );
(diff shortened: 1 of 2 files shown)


goldenapples:
I can see it being good to differentiate between term in different taxonomies with the same name, but I'm not sure about adding slugs in there. That seems like a very specialized case - in most situations, the content author shouldn't be concerned with the slugs for any terms.

Also, can you update this so that the taxonomy is only shown if more than one taxonomy is being queried, similarly to how the post type is being shown on `post_select` fields?

lvl99:
Thanks for your input @goldenapples.

I modified the post_select and term_select to show relative URLs instead of slugs (what I mainly needed was to see the location of where the page is, not just the slug).

I also made it show the relative URL and taxonomy type only if there are multiple post types/taxonomies configured for the field.

PR #776, opened 2017-09-29
Decoding of attributes not working 

sathyapulse:
When we set the encode attribute to any field, the values are encoded and stored in the database but the decode is not working in the post edit screen or the front end. The reason for that is the action `register_shortcode_ui` is called from https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L81 and it will hook the filter here https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L170. The filter is not called every time. It causes issues with decoding.

The PR does not actually solves the problem but it will enable developers to register the filter outside of the plugin. The decode function is not working when we hook it because it uses the private variable to get the shortcodes. The PR changes modifies it to use the function.

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -506,16 +506,15 @@ class Shortcode_UI {
 
 		// Get current shortcode tag from the current filter
 		// by stripping `shortcode_atts_` from start of string.
-		$shortcode_tag = substr( current_filter(), 15 );
-
-		if ( ! isset( $this->shortcodes[ $shortcode_tag ] ) ) {
+		$shortcode_tag  = substr( current_filter(), 15 );
+		$shortcode_args = $this->get_shortcode( $shortcode_tag );
+		if ( ! $shortcode_args ) {
 			return $out;
 		}
 
 		$fields = Shortcode_UI_Fields::get_instance()->get_fields();
-		$args   = $this->shortcodes[ $shortcode_tag ];
 
-		foreach ( $args['attrs'] as $attr ) {
+		foreach ( $shortcode_args['attrs'] as $attr ) {
 
 			$default = isset( $fields[ $attr['type'] ]['encode'] ) ? $fields[ $attr['type'] ]['encode'] : false;
 			$encoded = isset( $attr['encode'] ) ? $attr['encode'] : $default;


goldenapples:
I'm not sure I understand the problem you're describing.

Are you saying that `shortcode_atts()` isn't returning decoded data, even when it's called with the shortcode name as its third parameter?

I'd look at the way you're calling `shortcode_ui_register_for_shortcode()` for that shortcode. If you're only registering the shortcode UI in the admin, you're going to have problems using some of the enhancements Shortcake makes to the core shortcode model - the automagic decoding of encoded attributes is one example.

sathyapulse:
@goldenapples Yes, the attributes are not decoded even after passing shortcode name as its third parameter.

The function `shortcode_ui_register_for_shortcode()` is called inside the action hook `register_shortcode_ui`.

The function `filter_shortcode_atts_decode_encoded` cannot be called outside of the plugin. I'm trying to call the function like below
`add_filter( "shortcode_atts_shortcode_name", [ Shortcode_UI::get_instance(), 'filter_shortcode_atts_decode_encoded' ], 5, 3 );`

The `if` statement stops the execution when the filter is registered outside of the plugin
https://github.com/wp-shortcake/shortcake/blob/master/inc/class-shortcode-ui.php#L511

CC: @mattheu

goldenapples:
My question, though, is: how are you registering the shortcode UI? On what hook?

It's true that the `register_shortcode_ui` hook doesn't get called outside the admin, so if you register your shortcode UI on that hook, you won't get access to the attribute decoder on the front end. You could always register the shortcode UI earlier, maybe on `init`, and then the attribute decoder should work as you need it to.

Or am I misunderstanding your question?

sathiyamoorthy-venkatesan-nuk:
@goldenapples Yes, I understand what you mean. The shortcode UI function is called in the action hook `register_shortcode_ui` and I changed it to `init` the decoding works now. I feel this PR can be merged to master. The PR enables the function to be used outside of the plugin.

sathyapulse:
@goldenapples Yes, I understand what you mean. The shortcode UI function is called in the action hook `register_shortcode_ui` and I changed it to `init` the decoding works now. I feel this PR can be merged to master. The PR enables the function to be used outside of the plugin.

PR #783, opened 2017-11-30
Expose EditShortcodeForm View for reuse markup.

rahulsprajapati:
Expose EditShortcodeForm View for reuse markup. i.e., in repeater field.
issue: #773

PR #784, opened 2017-11-30
Fix phpcs Latest WPCS standard alignment issue.

rahulsprajapati:
Fix phpcs Latest WPCS standard alignment issue.
FYI: Used **phpcbf** for this.

### DIFF
--- a/inc/fields/class-shortcode-ui-field-user-select.php
+++ b/inc/fields/class-shortcode-ui-field-user-select.php
@@ -5,7 +5,7 @@ class Shortcode_UI_Field_User_Select {
 	private static $instance;
 
 	// All registered user fields.
-	private $user_fields  = array();
+	private $user_fields = array();
 
 	// Field Settings.
 	private $fields = array(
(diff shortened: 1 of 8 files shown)


inc/fields/class-shortcode-ui-field-user-select.php
```diff
@@ -34,11 +34,11 @@ public static function get_instance() {
...
 	private function setup_actions() {
 
-		add_filter( 'shortcode_ui_fields',             array( $this, 'filter_shortcode_ui_fields' ) );
-		add_action( 'enqueue_shortcode_ui',            array( $this, 'action_enqueue_shortcode_ui' ) );
+		add_filter( 'shortcode_ui_fields', array( $this, 'filter_shortcode_ui_fields' ) );
+		add_action( 'enqueue_shortcode_ui', array( $this, 'action_enqueue_shortcode_ui' ) );
 		add_action( 'wp_ajax_shortcode_ui_user_field', array( $this, 'action_wp_ajax_shortcode_ui_user_field' ) );
 		add_action( 'wp_ajax_shortcode_ui_user_field_preselect', array( $this, 'action_wp_ajax_shortcode_ui_user_field_preselect' ) );
-		add_action( 'shortcode_ui_loaded_editor',      array( $this, 'action_shortcode_ui_loaded_editor' ) );
+		add_action( 'shortcode_ui_loaded_editor', array( $this, 'action_shortcode_ui_loaded_editor' ) );
```
goldenapples, line 41: It would be nice to not lose the alignment here, but if this is what cs calls for, then `¯\_(ツ)_/¯`...

goldenapples (approved):

goldenapples:
Thanks! If you merge this into #783 I can approve that as well...

inc/fields/class-shortcode-ui-field-user-select.php
rahulsprajapati: yeah, I saw the nesting alignment on more then 3 dimensional array, that one also have same kind a cs.

Issue #790, opened 2018-01-27
Ensure that media frame is set to the correct state before rendering Insert Shortcode view


goldenapples:
Re-selecting the "Insert post element" menu item in the media frame after navigating to a different modal state causes an "Uncaught TypeError: Cannot read property 'get' of undefined" javascript error.

Steps to reproduce:

* Click the "Insert post element" button
* 'Insert Post Element' modal popup opens
* Select any option from the shortcode list
* Change your mind, select another item from the menu - ie 'Add media'
* Re-select Insert post Element
* The frame fails to render, and the JS error shown above is displayed in the console.

Probably related to #788, but I wanted to open as a separate issue (I have a fix for this, but <del>I'm not sure if it addresses #788 yet</del> it doesn't fix that issue yet.)

goldenapples:
Fixed by #791.

PR #791, opened 2018-01-27
Ensure media frame is set to the correct state before rendering

goldenapples:
This fixes the bug noted in #790 by ensuring that the media frame is set
to the correct state before rendering.

(Side note, it looks like this problem has always existed here, but
until recently, the media menu was hidden with CSS, so there was no way
to navigate out of the Insert Shortcode frame state...)


Kenshino (approved):
LGTM

PR #792, opened 2018-01-27
Fix some phpcs warnings

goldenapples:
Update the coding standards on the site to allow builds to pass (still needs some work, a lot has changed in wpcs since the last update here.)


goldenapples:
Merging this so that the build can pass on other PRs.

Issue #788, opened 2018-01-24
Cannot upload media after inserting post element


elliott-stocks:
The "Select Files" button becomes unresponsive when clicking on "Add Media" tab after inserting post element.

To reproduce -

1. Click "Insert Post Element"
2. Select an element and insert
3. Click back on "Insert Post Element"
4. Click on "Add Media"
5. Click on "Select Files"

Or you can click on Add Media button -> Insert Post element -> Add media to reproduce the same issue.

Drag and drop seems to work fine, but can no longer click the button to select files.

Tested with WP 4.9.2 and Shortcake 0.7.3.

A workaround is re-initialize the uploader on `MediaController.reset()` once the state is set to `insert`.

`this.frame.uploader.uploader.uploader.init();`

But I think there's a better way.

https://github.com/wp-shortcake/shortcake/blob/0440ceed210dd578c00024064a1df7186588cffd/js/src/controllers/media-controller.js#L68

goldenapples:
I've just recently come across this same bug (perhaps it's new in WP 4.9, but I'm not sure what specific change might have caused this).

I'll dig around a bit and see if I can find an appropriate place to reset the controller when switching between tabs in the media modal.

goldenapples:
I haven't found anything more elegant than your suggestion of re-initializing the uploader when resetting the media controller. Think that might have be added, as hack as it seems. Care to open a PR?

goldenapples:
Fixed in #793.

PR #793, opened 2018-01-28
Re-initialize uploader after resetting media controller

elliott-stocks:
Fixes issue with 'Select Files' button within the 'Upload Media' view, as discussed in #788.

Was hoping for something less hacky, but I too couldn't find a solution that worked.


goldenapples (approved):
👍 LGTM

I'll merge this and do some digging into core and see if there's a better way to address this there in the future.

PR #794, opened 2018-02-06
Avoid outputting duplicate underscore templates

goldenapples:
Many of the attribute fields in this plugin were outputting their templates on the `shortcode_ui_loaded_editor` hook, which fires every time an editor is created. As a result, duplicate templates were being created with identical IDs when an edit screen contained more than one WP_Editor.

This PR addresses that problem by moving the template enqueues to the `enqueue_shortcode_ui` hook, which only fires once per page load, and from that hook, adding the template output functions onto the admin_print_footer_scripts hook, so that they show up in a reasonable place rather than in the middle of the page markup.


goldenapples:
Closing this PR, as the underlying issue is fixed in a slightly better way in #807.

Issue #785, opened 2017-12-06
Deactivate method on Media Controller?


elliott-stocks:
I noticed this todo on the select2 field. But as far as I can tell, the destroying on select2 doesn't actually happen;

```
/**
 * Extending SUI Media Controller to hide Select2 UI Drop-Down when menu
 * changes in Meida modal
 * 1. going back/forth between different shortcakes (refresh)
 * 2. changing the menu in left column (deactivate)
 * 3. @TODO closing the modal.
 */
```

I need to be able to detect when the modal closes, to trigger an action.

This is the code I used -

```
var mediaController = sui.controllers.MediaController;
sui.controllers.MediaController = mediaController.extend({
	deactivate: function() {
		console.log( 'do something');
	}
});
```

However this doesn't work, am I possibly calling it to early?

I'm thinking would it not be easier to add deactivate method on the media controller, then trigger a hook?

On `MediaController` initialize -

```this.on( 'deactivate', this.deactivate, this );```

Then the function that triggers a hook -

```
deactivate: function() {
	var hookName = 'shortcode-ui.frame_deactivate';
	wp.shortcake.hooks.doAction( hookName, this );
}
```

I'm happy to open a PR for this, if you think it'd make sense? Would then resolve the select2 UI issue too.

goldenapples:
Does the [`render_destroy`](https://github.com/wp-shortcake/shortcake/blob/0440ceed210dd578c00024064a1df7186588cffd/js/src/views/media-frame.js#L95-L106) hook do what you want? It looks like it's not currently being called if the media modal is closed without inserting to the editor, but it probably should be.

elliott-stocks:
Thanks, @goldenapples. I'm using that already, but as you mentioned it isn't called on modal close.

Adding this hook to the `deactivate` method seems like it would work, thoughts?

goldenapples:
> Adding this hook to the deactivate method seems like it would work, thoughts?

Seems feasible. However, since we're overriding the core media controller, I'd be wary of overloading it too much. What do you think about attaching it in the one-time "close" callback [here](https://github.com/wp-shortcake/shortcake/blob/0440ceed210dd578c00024064a1df7186588cffd/js/src/controllers/media-controller.js#L102-L104), so that the hook only gets called in the shortcode UI frame and not in every media controller?

desaiuditd:
[...]

PR #797, opened 2018-03-14
Add media close hook

cyruscollier:
Addresses Issue #785.

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -142,18 +142,18 @@ var MediaController = wp.media.controller.State.extend({
 
 			this.frame.once( 'close', function() {
 				this.frame.mediaController.toggleSidebar( false );
-                /* Trigger render_closed */
-                /*
-                 * Action run after the shortcode overlay is closed.
-                 *
-                 * Called as `shortcode-ui.render_closed`.
-                 *
-                 * @param shortcodeModel (object)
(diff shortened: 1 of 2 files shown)


goldenapples (approved):
Looks good to me - thanks for the PR!

We have some other hooks - specifically `render_destroy` that I was considering suggesting that you reuse for this purpose. But its better not to break any existing usage, in case any developers are using the render_destroy hook and depending on its current (buggy) behavior of only being triggered when a shortcode is sent to the editor.

So I think this can be merged as is - `render_closed` is a clear hook name, and I'll add a note in the `render_destroy` docblock clarifying that this hook is only run on inserting a shortcode.

desaiuditd:
@goldenapples  What event would you recommend to use on updating existing shortcode ? Would it still be `render_closed`?

> So I think this can be merged as is - render_closed is a clear hook name, and I'll add a note in the render_destroy docblock clarifying that this hook is only run on inserting a shortcode.

desaiuditd:
Apparently, while digging in further, I realised that the `render_destroy` not being triggered on close, could be because of https://github.com/wp-shortcake/shortcake-bakery/issues/219

https://github.com/wp-shortcake/shortcake-bakery/pull/220 solves the issue and supersedes this PR. 

@goldenapples Thoughts?

goldenapples:
Thanks @cyruscollier for the PR and @desaiuditd for testing... Merging this PR now.

PR #800, opened 2018-09-26
Add allowClear property to select2 inputs

circlecube:
Add the select2 property `allowClear` to select2 input fields. Set default value to false (since it's currently not included, though it may be better as set to true). 

This fixes issue 610 https://github.com/wp-shortcake/shortcake/issues/610

### DIFF
--- a/js/build/shortcode-ui.js
+++ b/js/build/shortcode-ui.js
@@ -1856,7 +1856,10 @@ sui.views.editAttributeSelect2Field = sui.views.editAttributeField.extend( {
 	render: function() {
 
 		var self = this,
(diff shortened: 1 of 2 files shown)


goldenapples:
This looks good, but can you make the change in the [src file](https://github.com/wp-shortcake/shortcake/blob/5849e07f9199d9494730deabd4d945163e539fb5/js/src/views/select2-field.js#L94) and then run the grunt task? Thanks.

circlecube:
Ha, that would help. (facepalm) - There we go.

goldenapples:
:100: Merging, so this can get into the next release version.

PR #801, opened 2018-10-10
Fixed broken image urls in readme.md

awelles:


goldenapples:
Thank you @awelles! Since this README.md file is automatically generated from the readme.txt, I added a fix to the Grunt readme task configuration so that these screenshot URLs are generated correctly in the future. 🚀 merging this now...

Issue #806, opened 2018-12-15
Gutenberg compatibility issue


billerickson:
When Shortcode UI is active in WordPress 5.0 or higher, it triggers a php notice:

> PHP Notice: edit_form_advanced is deprecated since version 5.0.0! Use block_editor_meta_box_hidden_fields instead. This action is still supported in the classic editor, but is deprecated in the block editor.

This PHP notice then triggers Gutenberg to load in “Quirks Mode”, which leads to additional issues like metaboxes appearing on top of blocks in the editor. Shortcode UI must be deactivated to make the block editor usable.

![screenshot](https://user-images.githubusercontent.com/88897/49469040-186d3780-f7cc-11e8-808f-f3ca1329616e.jpg)

Here’s a detailed report on the issue: https://github.com/WordPress/gutenberg/issues/12598

kadamwhite:
I'm also seeing the same quirks mode issue apply on a site that isn't noticeably displaying that error; in our case, it's the Shortcake JS templates which are being output too early.
```html
<script type="text/html" id="tmpl-fusion-shortcake-field-attachment">
<div class="field-block shortcode-ui-field-attachment shortcode-ui-attribute-{{ data.attr }}">
  <label for="{{ data.attr }}">{{{ data.label }}}</label>
  <# if ( typeof data.description=='string' ) { #>
    <p class="description">{{{ data.description }}}</p>
    <# } #>
      <button id="{{ data.attr }}" class="shortcake-attachment-select button button-small add">{{ data.addButton }}</button>
      <div class="attachment-previews"></div>
</div>
</script>
```
and so on. I would expect these to get output within the `<head>` or `<body>` of the page, not before `<!doctype>`

kasparsd:
This appears to have been fixed in #807. Should this be marked as closed?

goldenapples:
Yup, closing now.

PR #807, opened 2019-01-16
Avoid outputting underscore templates before the opening <doctype> tag

goldenapples:
As noted in #806, the pattern used for many of the Shortcake attribute fields which output custom field templates is to output those templates on the `shortcode_ui_loaded_editor` hook. In WP 5.0 with the block editor active, this hook fires before the page output is begun, which causes these script templates to be spit out before the opening <doctype> tag, causing all kinds of ruckus with the browser layout.

Specifically, any output before the <doctype> declaration forces browsers into quirks mode, which renders the flex layout required for the Gutenberg editor incorrectly. In practive this was causing meta boxes to float awkwardly over editor content.

This fixes the problem by, instead of outputting the templates on the `shortcode_ui_loaded_editor` hook, instead attaching a hook to a later action to output these templates in the admin footer.

### DIFF
--- a/inc/fields/class-shortcode-ui-field-attachment.php
+++ b/inc/fields/class-shortcode-ui-field-attachment.php
@@ -89,7 +89,7 @@ class Shortcode_UI_Field_Attachment {
 	 */
 	public function action_shortcode_ui_loaded_editor() {
 		add_action( 'admin_print_footer_scripts', array( $this, 'output_templates' ) );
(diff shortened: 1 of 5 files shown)


kadamwhite (approved):
Patch looks great, and I can confirm this resolves the issue both in a reduced test case and in a much more complex codebase! Thanks for the quick fix!

PR #810, opened 2019-08-25
Allow url field to accept class attribute

LC43:
this will allow devs to assign classes to specific elements in the admin edit screens.

then just

```php

array(
	'label'       => esc_html__( 'URL', 'cake ),
	'attr'        => 'url-1',
	'type'        => 'url',
	'class'       => 'cake_main_url-1',
	'description' => esc_html__( 'Link to open when button is clicked', 'cake' ),
),
```


goldenapples:
I can see how this might be helpful. 

The only suggestion I have to use some escaping around the class name, so that we have a bit of protection against broken markup or xss. Changing this from `{{{data.class}}}` to `{{ data.class }}` would use underscore's escaping and prevent quotes passed in this parameter from breaking the output markup. If you make that change, I'd be 👍for merging it.

LC43:
hi @goldenapples , thanks for the suggestion!

I've changed the PR accordingly.

PR #817, opened 2019-11-26
added is_array() for multiple post types

### LINKED ISSUE #816: Countable warning when post types is a string
In `action_wp_ajax_shortcode_ui_post_field`, the assignment of `$is_multiple_post_types` throws a warning when post types is not an array. This is happening on a site with PHP 7.2. https://github.com/wp-shortcake/shortcake/blob/master/inc/fields/class-shortcode-ui-field-post-select.php#L148

Adding an `is_array()` check before counting the results will resolve this warning. I'll get a PR submitted for this.

petenelson:
Resolves #816

### DIFF
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -145,7 +145,7 @@ class Shortcode_UI_Field_Post_Select {
 
 		$query                  = new WP_Query( $query_args );
 		$post_types             = $query->get( 'post_type' );
-		$is_multiple_post_types = is_array( $post_types ) && count( $post_types ) > 1 || 'any' === $post_types;
+		$is_multiple_post_types = ( is_array( $post_types ) && count( $post_types ) > 1 ) || 'any' === $post_types;
 
 		foreach ( $query->posts as $post_id ) {
 			$post_type     = get_post_type( $post_id );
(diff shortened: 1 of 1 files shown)


inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -145,7 +145,7 @@ public function action_wp_ajax_shortcode_ui_post_field() {
 
 		$query                  = new WP_Query( $query_args );
 		$post_types             = $query->get( 'post_type' );
-		$is_multiple_post_types = count( $post_types ) > 1 || 'any' === $post_types;
+		$is_multiple_post_types = is_array( $post_types ) && count( $post_types ) > 1 || 'any' === $post_types;
```
goldenapples: I haven't tested this, but it looks like this would cause an order-of-precedence issue where the last part of the conditional would never be reached because "any" is not an array.

Can we add some grouping here so that it's easier to read and doesn't change the meaning of the conditional?

```suggestion
		$is_multiple_post_types = ( is_array( $post_types ) && count( $post_types ) > 1 ) || 'any' === $post_types;
```

goldenapples (changes requested):
This is good to fix, just left a suggestion to make the logic more clear and potentially avoid a logic bug.

eocreative-ethan:
This PR was created 4 years ago yet here I am still today battling with this same issue on a bang up-to-date install. 

Is there a reason this PR was never merged?

goldenapples:
> Is there a reason this PR was never merged?

Probably just that this project isn't actively maintained.

This is clear enough of a change and low-enough risk that I'm happy to merge and cut a new release. But, I'm not working on any classic editor projects to be able to give future changes the kind of testing they deserve. If anyone is interested in stepping up and maintaining the plugin, I'm willing to help transition it to new maintainers.

PR #820, opened 2019-12-05
VIP linter fixes

svandragt:
A couple of issues were highlighted by the wp linter bot.

Sanitizes unsafe $_GET['include'] use and defines $args as an empty array before using it.

Tested against `?include=12,23a`, `?include[]=12&include[]=23a` and without parameter.

### DIFF
--- a/inc/fields/class-shortcode-ui-field-post-select.php
+++ b/inc/fields/class-shortcode-ui-field-post-select.php
@@ -93,8 +93,13 @@ class Shortcode_UI_Field_Post_Select {
 		$nonce               = isset( $_GET['nonce'] ) ? sanitize_text_field( $_GET['nonce'] ) : null;
 		$requested_shortcode = isset( $_GET['shortcode'] ) ? sanitize_text_field( $_GET['shortcode'] ) : null;
 		$requested_attr      = isset( $_GET['attr'] ) ? sanitize_text_field( $_GET['attr'] ) : null;
-		$include             = isset( $_GET['include'] ) ? $_GET['include'] : array();
+		
 
+		$include = filter_input( INPUT_GET, 'include', FILTER_SANITIZE_NUMBER_INT, FILTER_REQUIRE_ARRAY );
+		if ( ! is_array( $include ) ) {
+			$include = (array) explode( ',', filter_input( INPUT_GET, 'include', FILTER_SANITIZE_STRING ) );
+		}
+		$include = array_map( 'intval', $include );
 
 		$response = array(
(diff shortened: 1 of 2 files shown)


inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -93,6 +93,8 @@ public function action_wp_ajax_shortcode_ui_post_field() {
 		$nonce               = isset( $_GET['nonce'] ) ? sanitize_text_field( $_GET['nonce'] ) : null;
 		$requested_shortcode = isset( $_GET['shortcode'] ) ? sanitize_text_field( $_GET['shortcode'] ) : null;
 		$requested_attr      = isset( $_GET['attr'] ) ? sanitize_text_field( $_GET['attr'] ) : null;
+		$include             = isset( $_GET['include'] ) ? $_GET['include'] : array();
```
goldenapples: Just to confirm - does this pass the VIP linter test or does it expect a sanitization function here?

goldenapples (approved):
Changes look good, just curious if this sanitization will pass the linter checks.

(Also it would be good to update the coding standards here so that they include VIP coding standards. I'll do that in a followup.)

svandragt:
@goldenapples Yes you are correct I thought it would be impossible to get rid of the one remaining warning but the code as it stands doesn't raise any linter issues. Would you mind reviewing? Thanks :)

inc/fields/class-shortcode-ui-field-post-select.php
```diff
@@ -94,6 +94,12 @@ public function action_wp_ajax_shortcode_ui_post_field() {
 		$requested_shortcode = isset( $_GET['shortcode'] ) ? sanitize_text_field( $_GET['shortcode'] ) : null;
 		$requested_attr      = isset( $_GET['attr'] ) ? sanitize_text_field( $_GET['attr'] ) : null;
 
+		$include = filter_input( INPUT_GET, 'include', FILTER_SANITIZE_NUMBER_INT, FILTER_REQUIRE_ARRAY );
```
goldenapples, line 97: I'm not sure I'm following what `FILTER_SANITIZE_NUMBER_INT` is doing here, since you map `intval` over the array a couple lines below. Is the sanitize_int filter just necessary to pass the lint rule?

svandragt: Yes this seems like an overzealous linter change really, will revisit this PR and see if it's still relevant.

PR #821, opened 2020-02-05
fix %s in string

kuliebiakin:
Implement fix for **%s Details** in Norwegian version because currently users see **%** instead of shortcode label.


goldenapples (approved):
Thanks for the fix!

PR #824, opened 2022-06-28
Handle post_type not being an array

srtfisher:
Ensure that `$post_types` is an array to prevent a fatal PHP (the fatal error applies to PHP 8):

```
Fatal error: Uncaught Error: count(): Argument #1 ($value) must be of type Countable|array, string given
in ...

Call stack:

Shortcode_UI_Field_Post_Select::action_wp_ajax_shortcode_ui_post_field()
wp-includes/class-wp-hook.php:307
WP_Hook::apply_filters()
wp-includes/class-wp-hook.php:331
WP_Hook::do_action()
wp-includes/plugin.php:476
do_action()
wp-admin/admin-ajax.php:187
Query Monitor
There has been a critical error on this website.
```

### DIFF
--- a/inc/class-shortcode-ui.php
+++ b/inc/class-shortcode-ui.php
@@ -6,6 +6,14 @@
 class Shortcode_UI {
 
 	/**
+	 * Plugin version.
+	 *
+	 * @access private
+	 * @var string
+	 */
+	private $plugin_version;
+
+	/**
 	 * Path to the plugin directory.
 	 *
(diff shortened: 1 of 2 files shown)


goldenapples (approved):
Do you want this to be merged? It seems like a reasonable fix for people using this on PHP 8.

srtfisher:
@goldenapples yes, it also handles a bug for PHP 8.2.

